PR #10, opened 2021-11-01
[CHORE] Package Bumps

jakerainis:
Opening this PR to to A) prevent someone else from having to do and 2) testing out the upstream setup from a forked repo.

This PR updates all[^1] outdated packages. Most of these packages are patches and minor bumps but there are a couple of bigger ones:
- Next v12 🎉 
- Directus RC100 (the price field didn't seem to break this time 👍 )

Given the nature of all these updates, please do pull it down and play for a few minutes before approving. 

[^1]: All packages with the exception of Pino and ESLint (which I believe are both in the same situation where another dependency we use hasn't been updated to support their latest major versions).


apps/web/next.config.js
```diff
@@ -10,6 +10,7 @@ module.exports = transpileModules(
     },
     webpack: (config) => {
       config.experiments = {
+        ...config.experiments,
```
jakerainis, line 13: Found that overwriting the experiments object was blowing away some of Next's optimizations under the hood and causing errors.

smonn (approved):
Thanks for doing this!

IanPirro (approved):
🎉

smonn:
After running this locally, I do get some changes in my `package-lock.json`. I'm on Node v16 and npm v8. Everything seems to run fine though!

jakerainis:
@smonn I'll sync with you tomorrow before merging this. I'm puzzled why our package-locks are in flux.

afraser:
Well I'm definitely waiting for this to be merged before forking/cloning.

afraser:
@jakerainis @IanPirro are you guys are running different versions of node/npm?

afraser:
@jakerainis you might want to update the engines directive in package.json to specify node 16

jakerainis:
@smonn I just pushed a new lock using node 16 and npm 7. To @afraser's point, do you want me to bump the node version directives in the package.json files?

smonn:
> are you guys are running different versions of node/npm?

We probably were, but will can lock it to Node v16+ and npm v7+.

afraser:
If you do, make sure you get the readme as well. (btw, the only reason I mentioned 16 is that Dallas mentioned it) I have no strong preference on which version, just that the engines directive should be more specific

afraser:
--oh, and .nvmrc

smonn (approved):

jakerainis:
@afraser @smonn https://github.com/deptagency/algomart/pull/10/commits/60aa86fae22028bab4ec35bd4a6d9260d33d2e68

I bumped the directives, regen'd the lock file, and updated the docs (🙏 @afraser).

The .nvmrc is at lts so I think that should be okay for the next 2 years.

PR #17, opened 2021-11-02
NFT Template unique code should be not-nullable

smonn:


jakerainis (approved):

smonn:
Not sure what happened in nfty-mart, locally when creating the CMS DB from scratch the unique_code field was not nullable and unique. Adding this migration to ensure that is the case though.

IanPirro (approved):

PR #27, opened 2021-11-02
Hide auction filters when auctions are filtered out

afraser:
The disabled state for the switch component is indistinguishable from the enabled state. This is better at least for now.
 -- I'd also be happy to take a swing at disabled state styles if people don't like show/hide in this context.
 
### Showing auctions

CleanShot 2021-11-02 at 16 24 33@2x

### Hiding auctions

CleanShot 2021-11-02 at 16 24 46@2x

### DIFF
--- a/apps/web/src/components/releases/release-filter-type.tsx
+++ b/apps/web/src/components/releases/release-filter-type.tsx
@@ -39,56 +39,57 @@ export default function ReleaseFilterType() {
           />
         </div>
       </div>
-
-      {state.showAuction && (<>
-        <div className={clsx(css.filterRow, css.filterHeader)}>
-          <Heading level={2}>{t('release:filters.Auction Status')}</Heading>
+      <div className={clsx(css.filterRow, css.filterHeader)}>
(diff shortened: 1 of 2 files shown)


IanPirro (commented):
Personally I like the show and hide, I think right now the disabled state --> 50% opacity.

afraser:
@IanPirro oh that's crazy, I didn't even notice the disabled state was semi-transparent. I think it was extra subtle since all the toggles on the auction group were disabled and they're visually separate from the rest. Maybe we try 25%

IanPirro:
@afraser I never really noticed because the implementations i'm working on are more dark themed and the toggle color is darker vs. the teal so it looks okay. 25% may be better though.

Screen Shot 2021-11-02 at 1 31 12 PM

Screen Shot 2021-11-02 at 1 31 50 PM

afraser:
@IanPirro yeah, I also noticed that 30% is fine for the green color I have but way too hard to see when it's switched to off (very light gray background)

jakerainis (approved):

IanPirro (approved):

PR #32, opened 2021-11-03
Remove TF_VAR_ prefix from github secrets

kevlarr:
Passing Terraform input variables in via environment variables requires a specific convention, namely that a variable called `my_variable` would be passed in via `TF_VAR_my_variable=... terraform apply`.

This convention was followed with the github secrets themselves, meaning the [deploy file](https://github.com/deptagency/algomart/blob/6f7f3f1cac2588c9d44c01447c01f397c53b3cd1/.github/workflows/deploy.yml#L126) had lines like...

```
TF_VAR_api_creator_passphrase: ${{ secrets.TF_VAR_api_creator_passphrase }}
```

Since all of the environment variables had to be manually mapped to pass them through to Terraform, following that convention was unnecessary. Furthermore, it introduces some confusion because creating a new Github Secret named `TF_VAR_api_creator_passphrase` _looks like_ it saves as `TF_VAR_API_CREATOR_PASSPHRASE` even though it's referenced as `secrets.TF_VAR_api_creator_passphrase` in the workflows themselves.

This PR updates the deploy workflow to expect secrets that no longer have the `TF_VAR_` prefix and also are all-caps, eg:

```
TF_VAR_api_creator_passphrase: ${{ secrets.API_CREATOR_PASSPHRASE }}
```

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -119,60 +119,60 @@ jobs:
           TF_VAR_project: ${{ secrets.GCP_PROJECT_ID }}
           TF_VAR_credentials: ${{ secrets.GCP_SERVICE_ACCOUNT_KEY }}
 
-          TF_VAR_algod_host: ${{ secrets.TF_VAR_algod_host }}
-          TF_VAR_algod_key: ${{ secrets.TF_VAR_algod_key }}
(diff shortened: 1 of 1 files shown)


smonn (approved):
👍

Issue #47, opened 2021-11-04
Error when viewing 1 pack per user when having other purchases


IanPirro:
## Overview

When viewing a pack that only allows one purchase per customer the user will encounter an error if they have purchased other packs, but not the one per customer pack they are trying to view.

## To reproduce

1. Create a new user
2. Create a pack that allows only one purchase per user
3. Purchase any other pack
4. Find the one per customer pack in releases and navigate to that pack.
5. Will encounter 400 error

## Expected behavior

Be able to view and purchase the pack

## Screenshots

<img width="1786" alt="Screen Shot 2021-11-04 at 12 18 01 PM" src="https://user-images.githubusercontent.com/781839/140406478-192ec10f-7501-4d0d-b695-51e80f9eeb9b.png">

## Desktop (if applicable):

- OS: OSX 11.1
- Browser: Chrome
- Version: 95.0.4638.54

PR #49, opened 2021-11-04
[#47] Error when viewing 1 pack per user when having other purchases

IanPirro:
Closes: #47

### DIFF
--- a/services/api/src/modules/packs/packs.service.ts
+++ b/services/api/src/modules/packs/packs.service.ts
@@ -339,6 +339,7 @@ export default class PacksService {
         templateIds.includes(id)
       )
 
+      // User doesn't own any of the packs in provided template IDs
       if (templateIdsForOwner.length === 0) {
         return {
           packs: [],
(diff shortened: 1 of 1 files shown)


jakerainis (approved):
Damn, good find!

services/api/src/modules/packs/packs.service.ts
```diff
@@ -332,16 +332,28 @@ export default class PacksService {
...
     }
 
+    // If templateIdsForOwner has matches use the intersecting ids. Otherwise use owner's pack IDs.
+    filter.id =
+      templateIdsForOwner.length > 0
+        ? { _in: templateIdsForOwner }
+        : { _in: packsIdsByOwnerId }
+
     // Find templates for the packs owned by user
     const { packs: templates } = await this.cms.findAllPacks({
```
IanPirro, line 358: Seemed previously without this checkout `{ _in: templateIdsForOwner }` =  `{ _in: [] }` and was causing an error

dallashuggins (approved):
Nice find indeed! Confirmed this fixed the error for me. 🙌

PR #51, opened 2021-11-04
Video fit and finish

afraser:
Set background of flip cards to black so varying image sizes don't accidentally show-through to the back side of the video (per feedback from Ian).

Also automatically play/pause on flip so we don't waste cpu and so we can support audio in the future if need-be.

https://user-images.githubusercontent.com/1179515/140423529-e742efe4-c23e-40a1-b8fb-a21e1ce6dc07.mp4
(opacity set to 50% so you can see the automatic play/pause on flip)


jakerainis (approved):

IanPirro (approved):
🎉

PR #52, opened 2021-11-05
JIT Minting

smonn:
This PR changes the minting model to be just-in-time instead of ahead-of-time. This means the checkout flow will be slower as it needs to wait for the NFT to be minted before the transfer can happen. The upside is that the initial cost to mint thousands/millions of NFTs is a non-issue as the cost is deferred until someone buys the NFT(s).


jakerainis (approved):
Sorry for the delayed review here, but looking great. So from what I can tell, we still generate packs and collectibles via tasks, but the actual minting of the collectibles themselves only happens when a pack is purchased or claimed, correct?

smonn:
> Sorry for the delayed review here, but looking great. So from what I can tell, we still generate packs and collectibles via tasks, but the actual minting of the collectibles themselves only happens when a pack is purchased or claimed, correct?

@jakerainis That is correct.

PR #56, opened 2021-11-05
Set log level always on directus user count

IanPirro:
Fixes situation where admin user wouldn't get created if LOG_LEVEL wasn't set due to an additional log statement.

Screen Shot 2021-11-05 at 2 36 01 PM


jakerainis (approved):
Thanks for doing this 🙏

PR #61, opened 2021-11-09
exporting & importing CMS data

dallashuggins:
### Changes
- Adds file structure and functionality for importing CSV data into the CMS for a collection. Running `npm run import-create-starters` will output XLS documents that can be edited and exported as a CSV file.
- New library for parsing csv data.
- Updated `README.md` with instructions.


apps/cms/README.md
```diff
@@ -181,3 +181,25 @@ The homepage of the provided UI depicts a featured Pack, which takes prominence
...
+
+Run `npm run import-create-starters` to output documents for each collection inside the `scripts/import-data/starters/` folder. These XLS documents will include all fields for the collection as columns (the field ID is the column header i.e., `preview_image`). There must be a column for every field associated with the collection, but only required fields must have data, otherwise it can be left empty.
+
+Open these documents in the editor of your choosing. Fill out information for each of the items you'd like to import per collection.
+
+For any image fields, you can either specify a UUID for an existing photo in the CMS OR specify a filename (name + extension). The filename would reference a file added to the `scripts/import-data/images` folder. You will be prompted to provide the column header (i.e., the field ID) for each of the image fields, so that it will know to check for any filenames for upload.
+
+If the status is not provided, it will be assigned the 'draft' status.
+
+2. Place new CSV file inside the `scripts/import-data` folder.
```
afraser, line 199: Is the thinking here that these files be added to source control so the data can be synced across devs?

afraser: ...just noticed the gitignore. IMO it'd be better to either commit this file or simply provide it on the command line (https://nodejs.org/en/knowledge/command-line/how-to-parse-command-line-arguments/)

smonn (approved):

apps/cms/README.md
```diff
@@ -181,3 +181,25 @@ The homepage of the provided UI depicts a featured Pack, which takes prominence
...
+
+After filling out the document(s) with any items you'd like to import, export the file as a CSV into the `scripts/import-data` folder. The filename must follow the structure: `collection].csv`.
```
jakerainis, line 201:
```suggestion
After filling out the document(s) with any items you'd like to import, export the file as a CSV into the `scripts/import-data` folder. The filename must follow the structure: `[collection].csv`.
```
Looks like an unclosed bracket?

afraser (approved):
Love it.

afraser (changes requested):
I love it. My only issue is I don't think the files being output by this are valid XLS files, but they _are_ valid CSVs. I think we should probably use that extension instead.

dallashuggins:
> I love it. My only issue is I don't think the files being output by this are valid XLS files, but they _are_ valid CSVs. I think we should probably use that extension instead.

Yay! Ah yeah, you're right. If you open with a 

> I love it. My only issue is I don't think the files being output by this are valid XLS files, but they _are_ valid CSVs. I think we should probably use that extension instead.

@afraser Ah yeah, I think you're right. I realized opening a CSV auto-opens Numbers for me, so we don't need to open XLS documents. Changing to use csv.

apps/cms/scripts/import-data/import.mjs
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Begin importing data.
+   */
+  console.log('Starting import...')
+  const basePath = './scripts/import-data'
+  const files = await groupFilesFromDirectoryByExtension(basePath, 'csv')
+  // Loop through the import data
+  for (const file of files) {
+    const csvFile = _resolve(`${basePath}/${file}`)
```
IanPirro, line 44: super minor, but if the order doesn't matter you could probably

```javascript
const processFile = async (file, basePath, token) => {
  const csvFile = _resolve(`${basePath}/${file}`)
  const collection = file.split('.').shift()
  console.log(`Checking file data for ${collection}...`)
  // Parse CSV and check values.
  const csvData = await parseCsvData(csvFile)
  await checkCsvAsync(csvData, collection, token)
  // Create images and update file.
  const data = await updateCsvAsync(csvData, basePath, token)
  const formData = new FormData()
  const jsonFile = `${basePath}/${collection}.json`
  // Create JSON file.
  writeFileSync(jsonFile, JSON.stringify(data))
  formData.append('file', createReadStream(jsonFile))
  // Import data to CMS
  console.log(`Importing file for ${collection}...`)
  await importDataFile(formData, collection, token)
  // Remove newly created JSON file
  await removeFile(jsonFile)
}
```

```javascript
const promises = []
  // Loop through the import data
  for (const file of files) {
    promises.push(processFile(file, basePath, token))
  }
  await Promise.all(promises)

  console.log('Done!')
```

to slightly speed it up

afraser (approved):

apps/cms/scripts/import-data/import.mjs
dallashuggins: Good idea! Currently there's no order except alphabetical with the CSV files so this would work great. Long-term it'd be great if we could import in order and reference new items being created in the import process, but there would be some complexity there. Going to test a bit but for the current setup this should be perfect.

dallashuggins: Ah actually found an issue. It tries to execute the readlineAsync function at the same time, so as you type image values it writes every letter multiple times.

smonn (commented):
The `apps/cms/package-lock.json` can be deleted. Only the root `package-lock.json` should exist. May wanna add a line in the `.gitignore` to avoid adding these going forward, something like this:

```
**/package-lock.json
!package-lock.json
```

jakerainis (approved):

Issue #76, opened 2021-11-15
Scheduled task for checking wire payments


dallashuggins:
## Overview

Create task for checking the Circle payments endpoint for successful wire transfer. 

We cannot transfer the asset automatically without the passcode, so this may be for sending them an email to let them know payment was verified, and then they can login to make the transfer. We'll need a form for gathering the passcode and initiating the transfer, since payment will be complete.

## Additional context

We currently have a task for retrieving pending payments and looking up the payment status and updating it in the database. There will not be a payment created initially, so this cannot be used for checking for wire transfers.

The [get payments endpoint](https://developers.circle.com/reference#payments-payments-get) has a query param for filtering by wire transfers (`type` = `wire`). The `source.id` attribute matches the ID of the bank account initially created. The status will be `paid` when the funds have settled.

This could be a manual step though, so this isn't necessary but a nice-to-have. But we'll need to mark the user as the owner of the pack once we get confirmation of payment.

Should there be a time limit to how long the user has to initiate wire transfer? AKA should we support bid expiration for these auctions. If we do, we should probably double check the status of payment right before assigning to next bidder, and maybe add time if the status is pending. 

We need to check the price of the successful payment as well during this process.

CarterRose:
Hey text me if you need code for transfer

dallashuggins:
Closing since the ability to find a wire payment using the scheduled task was completed in #75 and the ticket #138 is for handling the paid status, though we may just not allow transfer out but award the pack right away, and set an expiration date for monitoring for payment.

PR #79, opened 2021-11-16
[#75] Accepting wire payments

### LINKED ISSUE #75: Accepting wire payments
## Overview

This is for building out the functionality to support accepting wire payments.

## Additional context

A lot of the functionality was built for another project and just needs to be pulled into this project.

_Including_
- API endpoints to support front-end functionality (creating bank accounts, getting the status, and getting wire instructions) and corresponding types.
- Front-end endpoints and validation.
- Simplify bid flow for wire payments (no need to gather credit car information)
- Checkout flow for initiating creating an account for the wire payment

dallashuggins:
### Changes
- API endpoints to support bank account creation for wire transfers, and for getting the bank account status and instructions for the wire payment. Corresponding endpoints in web.
- If the current bid or pack price is higher than $3000, the wire account flow will be used instead of the payment/card flow. 
  - This includes not requiring card information on the submission of the form, only a checkbox (currently says "I promise to pay", but we should update that) on bid submission, and then initiating creating the wire bank account in Circle. 
  - The user then takes the information from the success page and initiates payment.
  - An email is sent to the user upon successful bank account creation. 
- Feature flag `WIRE_PAYMENT_ENABLED` to enable/disable if the bank account component would be displayed, which is what would handle the bank account creation to initiate the wire transfer flow.

### Examples

Screen Shot 2021-12-03 at 12 49 39 PM

Closes #75 
Progress on #76 
* Last thing: how to handle not awarding a pack before a wire is received


apps/web/src/components/payment-methods-form/index.tsx
```diff
@@ -8,13 +8,22 @@ import AddMethodsSuccess from './sections/add-methods-success'
...
 export interface MyProfilePaymentMethodsAddProps {
   formErrors?: ExtractError<
-    ReturnType<typeof validatePurchaseForm | typeof validateExpirationDate>
+    ReturnType<
+      | typeof validateBankAccount
+      | typeof validateBidsForm
+      | typeof validateBidsFormForWires
+      | typeof validatePurchaseForm
+      | typeof validateExpirationDate
+    >
```
dallashuggins, line 26: The changes in unrelated files are because we're sharing the `formErrors` validation.

services/api/src/modules/payments/payments.service.ts
```diff
@@ -89,6 +91,44 @@ export default class PaymentsService {
...
+    }
+  }
+
+  async getWireTransferInstructions(bankAccountId: string) {
+    // Find bank account in database
+    const foundBankAccount = await PaymentBankAccountModel.query().findById(
+      bankAccountId
+    )
+
+    // If a bank account was found, use the external ID. Otherwise check by passed in ID.
```
dallashuggins, line 119:
```suggestion
```

services/api/src/modules/payments/payments.service.ts
```diff
@@ -89,6 +91,44 @@ export default class PaymentsService {
     return cards
   }
 
+  async getBankAccountStatus(bankAccountId: string) {
+    // Find bank account in database
+    const foundBankAccount = await PaymentBankAccountModel.query().findById(
+      bankAccountId
+    )
+
+    // If a bank account was found, use the external ID. Otherwise check by passed in ID.
```
dallashuggins, line 100:
```suggestion
```

apps/web/src/environment.ts
```diff
@@ -64,4 +64,19 @@ export const Environment = {
...
+    )
+    switch (isEnabled) {
+      case 'true':
+        return true
+      case 'false':
+        return false
+      default:
+        return isEnabled
+    }
+  },
```
dallashuggins, line 81: I added this switch because I was not able to get a boolean to return for the env variable, it was always a string.

smonn: Right, I guess the `this.config` method is a bit misleading. Can remove the generic typing from it and have it always return a string, and then in cases like this where we need some other type handle it accordingly. The Firebase configs and the currency one are two examples of need something other than a string.

dallashuggins: `JSON.parse` seems to always return the right value as well. Was thinking of changing to: ```return typeof JSON.parse(isEnabled) === boolean ? JSON.parse(isEnabled) : false``` Do you see any issue with this?

smonn (approved):

apps/web/src/environment.ts
dallashuggins: Just committed your suggestion, which is nice and clean. Thanks!

smonn: Potential issue with using `JSON.parse` is that if it's an empty string it will throw an error.

services/api/src/models/payment-bank-account.model.ts
```diff
@@ -0,0 +1,29 @@
...
+import { BaseModel } from './base.model'
+import { UserAccountModel } from './user-account.model'
+
+export class PaymentBankAccountModel extends BaseModel {
+  static tableName = 'PaymentBankAccount'
+  static jsonSchema = PaymentBankAccountSchema
+
+  ownerId!: string
+  externalId!: string
+  status!: PaymentBankAccountStatus | null
```
smonn, line 16: The `status` column in the migration is marked as not nullable. So you'll wanna remove `| null` here or remove the `.notNullable()` from the migration.

```suggestion
  status!: PaymentBankAccountStatus
```

dallashuggins: Nice find, thanks!

jakerainis (approved):

services/api/src/migrations/20211116140144_payment_bank_account.ts
```diff
@@ -0,0 +1,26 @@
...
+    table.text('description').notNullable()
+    table.uuid('ownerId').notNullable().references('id').inTable('UserAccount')
+    table.timestamp('createdAt').defaultTo(knex.fn.now()).notNullable()
+    table.timestamp('updatedAt').defaultTo(knex.fn.now()).notNullable()
+  })
+  await knex.schema.alterTable('Payment', (table) => {
+    table.uuid('paymentBankId').references('id').inTable('PaymentBankAccount')
+    table.uuid('externalId').nullable().alter()
+  })
+}
```
dallashuggins, line 18: The `Payment` table needs to allow null values for the external ID, so that a payment can be created for the wire transfer before the payment is available via Circle's API. This links the payment as well as the pack to the wire transfer. When the scheduler function is added, we can update the payment with the status and if it fails, remove the claim to the pack.

services/api/src/modules/payments/payments.service.ts
```diff
@@ -315,6 +459,25 @@ export default class PaymentsService {
...
+    })
+    if (!payments) return null
+    // Find payment with matching source ID
+    const sourcePayments = payments.map(
+      (payment) => payment.source.id === sourceId
+    )
+    userInvariant(sourcePayments, 'payments not found', 404)
+    return sourcePayments
+  }
+
```
dallashuggins, line 480: For checking for the presence of a matching payment in Circle - will be used in the upcoming scheduler task

services/api/src/modules/payments/payments.service.ts
```diff
@@ -351,20 +588,23 @@ export default class PaymentsService {
...
+            circlePayment,
+            `external payment ${payment.externalId} not found`
+          )
+
+          if (payment.status !== circlePayment.status) {
+            await PaymentModel.query(trx).patchAndFetchById(payment.id, {
+              status: circlePayment.status,
+            })
+            updatedPayments++
+          }
```
dallashuggins, line 605: Since `externalId` can be null, this conditional was needed.

services/api/src/modules/payments/payments.service.ts
```diff
@@ -149,6 +193,92 @@ export default class PaymentsService {
...
+
+    // Create new payment in database
+    const newPayment = await PaymentModel.query(trx).insert({
+      externalId: null,
+      payerId: user.id,
+      packId: packId,
+      paymentBankId: newBankAccount.id,
+      paymentCardId: null,
+      status: PaymentStatus.Pending,
+    })
```
dallashuggins, line 251: Adding a payment now during the bank account creation.

apps/web/src/components/bank-account-form/sections/bank-account-form.tsx
```diff
@@ -0,0 +1,244 @@
...
+                    : '',
+                country:
+                  formErrors && 'country' in formErrors
+                    ? (formErrors.country as string)
+                    : '',
+                zipCode:
+                  formErrors && 'zipCode' in formErrors
+                    ? (formErrors.zipCode as string)
+                    : '',
+              }}
```
smonn, line 152: Could we align the types here and in other places (for `formErrors`) to avoid the repeated checks and casting?

apps/web/src/components/bank-account-form/sections/bank-account-success.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  const isActiveAuction =
+    release.type === PackType.Auction &&
+    isAfterNow(new Date(release.auctionUntil as string))
+
+  const handleReturnToListing = useCallback(() => {
+    const path = urls.release.replace(':packSlug', release.slug)
+    if (typeof window !== 'undefined') {
+      window.location.assign(new URL(path, window.location.origin).href)
+    }
```
smonn, line 36: Is using `router.push(path)` not viable here? (where `const router = useRouter()`)

dallashuggins: Good call! Changed in the other branch with [fd3d2bd](https://github.com/deptagency/algomart/pull/123/commits/fd3d2bd79ade44c23c4d081fd5b9d37e269a3de5)

apps/web/src/components/bank-account-form/sections/bank-account-form.tsx
dallashuggins: Did you have an idea how to do this? I can look into it more, definitely would be ideal, but since we're group the different validations this is the only way I know to make it not throw an error saying the property doesn't exist.

smonn (approved):

afraser:
Small thing but the "success" messaging might give people the wrong idea (ie: that they're done and that the following text is just informational). Should we remove that?

dallashuggins:
> Small thing but the "success" messaging might give people the wrong idea (ie: that they're done and that the following text is just informational). Should we remove that?

@afraser Good point! Since this is feature flagged I could add in a future PR if we wanted to get this in, or do similar to the other PR and merge into this PR, but I think I'll update the success screen and general style according to the design Tom worked on.

apps/web/src/utils/purchase-validation.ts
```diff
@@ -8,12 +8,17 @@ import {
...
   object,
+  oneOf,
   required,
   string,
 } from 'validator-fns'
 
 import { formatCurrency } from '@/utils/format-currency'
 
+// Maximum bid for card payments as integer
+// eslint-disable-next-line unicorn/numeric-separators-style
```
dallashuggins, line 19: Throws the below error with `3_000_00` and this is an integer, so the disable line is needed. 
```
Invalid group length in numeric value.eslintunicorn/numeric-separators-style
```

dallashuggins:
@afraser I did end up removing "Success" in this PR and making the style closer to the designs Tom made.

apps/web/src/components/checkbox.tsx
```diff
@@ -2,30 +2,33 @@ import clsx from 'clsx'
...
+            'bg-gray-200': readOnly || disabled,
+          },
+          className
+        )}
+        checked={checked}
+        disabled={disabled}
+        readOnly={readOnly}
+        {...props}
+      />
+      {label && <span className="pl-2">{label}</span>}
```
afraser, line 31: nice!

apps/web/src/components/bank-account-form/sections/bank-account-success.tsx
```diff
@@ -38,28 +38,19 @@ export default function BankAccountSuccess({
   return (
     <div className={css.root}>
       <CheckCircleIcon className={css.icon} height="48" width="48" />
-      {release.type === PackType.Auction && (
+      {release.type === PackType.Auction && isActiveAuction && (
```
afraser, line 45: nice!

afraser (approved):

apps/web/src/components/bid-activity/bid-activity.tsx
```diff
@@ -113,7 +112,7 @@ export default function BidActivity({
                 )}
               </BidActivityDetails>
               {meetsReservePrice &&
-                (!isFollowingBid || followingBidDoesNotMeetReservePrice) && (
+                (!followingBid || followingBidDoesNotMeetReservePrice) && (
```
dallashuggins, line 115: The "Reserve price met" activity was never being displayed, since the reserve price (price from the CMS) is an integer, and we were converting to an integer again and therefore the price was always higher.

PR #80, opened 2021-11-16
Delay initial balance

### LINKED ISSUE #62: Defer sending initial funds until first NFT transfer
## Overview

To avoid locking up Algos in unused accounts (0.1 Algo per account), we should defer sending the initial amount until a user acquires their first NFT. This will ensure we keep funds in the funding wallet for as long as possible and is great for storefronts that will mostly rely on auctions where the NFT volume is relatively low.

smonn:
Closes #62

### DIFF
--- a/packages/schemas/src/accounts.ts
+++ b/packages/schemas/src/accounts.ts
@@ -12,7 +12,7 @@ export const AlgorandAccountSchema = Type.Intersect([
       minLength: 58,
       maxLength: 58,
     }),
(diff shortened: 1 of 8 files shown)


dallashuggins (approved):
Awesome! 🙌 Confirmed I could create an account and it didn't add the creation transaction ID to the new account, and then purchased a pack and it added the transaction ID successfully.

IanPirro (approved):
Also tried it out locally, LGTM :shipit: 🎉

PR #81, opened 2021-11-17
export cms data script

afraser:


apps/cms/scripts/export-data/export.mjs
```diff
@@ -0,0 +1,62 @@
...
+  console.log(`Creating export directory at ${exportDir}`)
+  if (!fs.existsSync(exportDir)) {
+    fs.mkdirSync(exportDir)
+  }
+
+  // Create a file for each collection
+  const collections = await getCollections(token)
+  await Promise.all(collections.map(collection => exportCollectionToCsv(exportDir, collection, token)))
+  console.timeEnd('Export')
+}
```
dallashuggins, line 56: Works great! 👏

dallashuggins: One thought is the image IDs exported won't match an image if it's being imported into a new db. May be worth documenting.

dallashuggins (approved):
The export works great, nice improvements! 🔥 

Side note, I realized I didn't take out the `Promise.all` for the import in my branch, so the import doesn't work as it should (when typing during the image directive, it repeats each key press), so I need to update that. Will do that after we get this merged in.

PR #91, opened 2021-11-18
[MISC] Pack Opening improvements & TS Node Fix

jakerainis:
## Overview:
- Adjust layouts of revealed cards in pack so that if there's one or two, they're prominently centered. If there's more, it defaults to the grid layout. Also adjusts some of the styles in the cards so the sizes aren't stretched undesirably.
- Adds `hideContent` prop to pack grid. When true, it only shows the question mark, and not the pack metadata, which is arguably only relevant in packs that have many cards. Right now, this shows content if there are 3 or more.
- Also added the "View In My Collection" button from the designs.
- Ran into an issue in DRL implementation (which uses some more up-to-date packages) where the API's knex migrations will fail since they're triggered using the node binary. This PR adds TS Node to ensure it compiles.

## Screenshots:
screencapture-localhost-3000-pack-opening-ec05b818-e631-4fa9-9dbe-ad9f888a3161-2021-11-18-14_23_55

screencapture-localhost-3000-pack-opening-ec05b818-e631-4fa9-9dbe-ad9f888a3161-2021-11-18-14_24_57

screencapture-localhost-3000-pack-opening-ec05b818-e631-4fa9-9dbe-ad9f888a3161-2021-11-18-14_25_35

### DIFF
--- a/apps/web/locales/en-US/common.json
+++ b/apps/web/locales/en-US/common.json
@@ -45,6 +45,7 @@
     "Sign Out": "Sign Out",
     "Try Again": "Try Again",
     "View All Payment Methods": "View All Payment Methods",
+    "View In My Collection": "View In My Collection",
     "View My Collection": "View My Collection",
     "View On AlgoExplorer": "View on AlgoExplorer",
     "View Release": "View Release",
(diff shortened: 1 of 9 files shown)


smonn (approved):
👍

IanPirro (approved):
🎉

dallashuggins (approved):
🔥

PR #109, opened 2021-11-22
Fallback UI for untransferred packs

### LINKED ISSUE #85: Trigger transfer after package opening
## Overview

Change the checkout flow so that it only deals with payments and minting of assets. The asset transfer will be moved to after the pack opening screen. This is somewhat related to the wire payment epic #74 

## Features

- [x] #62
- [x] #87 
- [x] #86 
- [x] #88 
- [x] #89 

### Additional context

This way the checkout will speed up a bit and the minting process can happen in the background while the user opens their pack. We'll still need to wait for the pack's NFTs to be minted before triggering the transfer. This also means moving the passphrase entry to show after the pack opening, most likely in a modal (similar to how it shows for free packs today).

May need some design input here too as we'll need to show in the UI, where appropriate, that NFTs are still being minted. We'll also need to provide an option to transfer a pack on a user's list of transactions (i.e. `/my/profile/transactions`).

#### Passphrase prompt

image

#### New transfer button needed on these screens

_/my/profile/transactions_

image

_/pack-opening/:id_

image

#### Flowchart for payment checkout

AlgoMart Payment Checkout - Payment Checkout

#### Flowchart for transfer

AlgoMart Payment Checkout - Transfer

### LINKED ISSUE #89: UI updates to resolve untransferred packs
## Overview

This will look and work similar to the email verification banner (see screenshot), except that it will only show if you have one or more packs waiting to be transferred. This can happen if you miss the "Save to My Collection" button after opening a pack. The CTA should take you to the related pack, and from there you should be able to trigger the transfer flow for that pack, assuming you own one and haven't transferred it. This new UI on the pack page would replace the "Purchase" CTAs for purchasable packs and the "Claim NFT" for won auction packs.

image

smonn:
Closes #89, closes #85

- fix email verification name typo
- extract banner component
- add endpoint to fetch untransferred packs
- initial stab at fetching and transferring untransferred packs

### DIFF
--- a/apps/web/src/layouts/default-layout.tsx
+++ b/apps/web/src/layouts/default-layout.tsx
@@ -6,7 +6,7 @@ import AppFooter from '@/components/app-footer/app-footer'
 import AppHeader from '@/components/app-header/app-header'
 import HeadTag from '@/components/head-tag/head-tag'
 import MainPanel from '@/components/main-panel/main-panel'
-import EmailVerfication from '@/components/profile/email-verification'
+import EmailVerification from '@/components/profile/email-verification'
 
 export interface DefaultLayoutProps {
   children?: ReactNode
@@ -29,7 +29,7 @@ export default function DefaultLayout({
     <>
       <HeadTag pageDescription={pageDescription} pageTitle={pageTitle} />
       <div className={common.backgroundGradient}>
-        <EmailVerfication />
+        <EmailVerification />
 
         <section>
           <AppHeader />
--- a/apps/web/src/layouts/default-layout.tsx
+++ b/apps/web/src/layouts/default-layout.tsx
@@ -7,6 +7,7 @@ import AppHeader from '@/components/app-header/app-header'
 import HeadTag from '@/components/head-tag/head-tag'
 import MainPanel from '@/components/main-panel/main-panel'
 import EmailVerification from '@/components/profile/email-verification'
+import UntransferredPacks from '@/components/profile/untransferred-packs'
 
 export interface DefaultLayoutProps {
   children?: ReactNode
@@ -30,6 +31,7 @@ export default function DefaultLayout({
       <HeadTag pageDescription={pageDescription} pageTitle={pageTitle} />
       <div className={common.backgroundGradient}>
         <EmailVerification />
+        <UntransferredPacks />
 
         <section>
           <AppHeader />
--- a/packages/schemas/src/packs.ts
+++ b/packages/schemas/src/packs.ts
@@ -120,6 +120,7 @@ export const PublishedPacksSchema = Type.Object({
 export const PackByOwnerSchema = Type.Intersect([
   PackBaseSchema,
   Type.Object({
+    id: IdSchema,
     activeBid: Type.Optional(Type.Number()),
     claimedAt: Type.String({ format: 'date-time' }),
   }),
--- a/services/api/src/modules/packs/packs.service.ts
+++ b/services/api/src/modules/packs/packs.service.ts
@@ -7,6 +7,7 @@ import {
   DEFAULT_LOCALE,
   EventAction,
   EventEntityType,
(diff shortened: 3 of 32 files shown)


services/api/src/modules/packs/packs.service.ts
```diff
@@ -737,6 +739,59 @@ export default class PacksService {
...
 
+  async untransferredPacks({
+    externalId,
+    locale = DEFAULT_LOCALE,
+  }: LocaleAndExternalId) {
+    const packs = await PackModel.query()
+      .withGraphJoined('[collectibles, owner]')
+      .where('owner.externalId', externalId)
+      .whereNull('collectibles.ownerId')
+      .whereNotNull('collectibles.address')
```
smonn, line 750: Exclude unminted collectibles, for now.

packages/schemas/src/packs.ts
```diff
@@ -120,6 +120,7 @@ export const PublishedPacksSchema = Type.Object({
 export const PackByOwnerSchema = Type.Intersect([
   PackBaseSchema,
   Type.Object({
+    id: IdSchema,
```
smonn, line 123: Need the pack `id` to be able to initiate a transfer.

apps/web/src/layouts/default-layout.tsx
```diff
@@ -29,7 +30,8 @@ export default function DefaultLayout({
     <>
       <HeadTag pageDescription={pageDescription} pageTitle={pageTitle} />
       <div className={common.backgroundGradient}>
-        <EmailVerfication />
+        <EmailVerification />
+        <UntransferredPacks />
```
smonn, line 34: May need to make these mutually exclusive.

services/api/src/modules/packs/packs.service.ts
jakerainis: Theoretically, the minting will have occurred in most cases by this point in most cases. But in this particular scenario, the user will be re-prompted again once unminted assets they've claimed are finally minted?

apps/web/src/layouts/default-layout.tsx
smonn: Or perhaps not? You need to have a verified email address before you can purchase a pack, right? @dallashuggins / @jakerainis

dallashuggins: Ah yes, that is correct, email verification is required to purchase a pack. So I don't think there'd be a case where someone would get both alerts.

services/api/src/modules/packs/packs.service.ts
smonn: A user shouldn't end up in a situation where this is hit and finds unminted packs. It happened to me locally as I was working on this feature where the minting had failed. Can chat a bit more about how we wanna handle this after standup today.

dallashuggins (approved):
Works great! 🔥

services/api/src/modules/packs/packs.service.ts
smonn: Pretty much. There are two cases the unminted ones could be found with this query (and as a result need to be excluded):

1. This endpoint is hit just after they've purchased a pack, but before the minting has completed. For this, excluding them makes sense.
2. If you purchase a pack but we fail to start the minting process or some error occurs that makes it impossible to finish the minting process. Then those packs' minting processes will need to be retried. Today, we have no mechanism to handle that and I don't think it should be part of this.

So the question is, what do we do in the latter case?

IanPirro (approved):

services/api/src/modules/packs/packs.service.ts
dallashuggins: Ah ok gotcha. Hmm good question. I can see having a task that looks for completed payments without minted assets, but wondering how we'd avoid minting collectibles when the process had been started but was not complete. 🤔

PR #114, opened 2021-11-24
Use nodemailer if SMTP is configured

### LINKED ISSUE #30: Switch to nodemailer to support SMTP and reduce vendor lock-in
## Overview

Right now the API requires a SendGrid account. To remove that vendor lock-in, we can switch to using [Nodemailer](https://nodemailer.com/about/) and SMTP configuration. SendGrid, among many other email service providers, [support using SMTP](https://docs.sendgrid.com/for-developers/sending-email/integrating-with-the-smtp-api).

Then we can use SMTP "connection strings" like these:

```
SMTP_URL=smtp://user:password@host:port/?option=a
```

And later in code:

```ts
import nodemailer from "nodemailer";

const transporter = nodemailer.createTransport(process.env.SMTP_URL);

transporter
  .sendMail({
    from: "sender@test.local",
    to: "recipient@test.local",
    subject: "My subject line",
    html: "<p>Hello there</p>",
  })
  .then((result) => {
    console.log(result);
  })
  .catch((error) => {
    console.log(error);
  });
```

smonn:
Closes #30

Allows the use of SMTP via nodemailer.

You can still use SendGrid as before with the same env vars, but this PR introduces a few new options:

```env
# Email configuration

# Sets the from email address and name
EMAIL_FROM=
EMAIL_NAME=AlgoMart

# Supported transports: smtp, sendgrid
EMAIL_TRANSPORT=smtp

# Required if EMAIL_TRANSPORT=sendgrid
SENDGRID_API_KEY=

# Required if EMAIL_TRANSPORT=smtp
# Can use services like https://ethereal.email/ or https://mailtrap.io/ during development
SMTP_HOST=
SMTP_PORT=
SMTP_USER=
SMTP_PASSWORD=
```

`EMAIL_FROM` replaces `SENDGRID_FROM_EMAIL`, but you can still use the latter as it's configured as a fallback. Also note the new `EMAIL_NAME` which sets the name of the sender.


services/api/src/lib/i18n-adapter.ts
```diff
@@ -0,0 +1,30 @@
...
+      lng: DEFAULT_LOCALE,
+      ns: ['emails'],
+      backend: {
+        loadPath: path.join(__dirname, '../../locales/{{lng}}/{{ns}}.json'),
+        addPath: path.join(
+          __dirname,
+          '../../locales/{{lng}}/{{ns}}.missing.json'
+        ),
+      },
+    })
```
smonn, line 20: Set this up to more closely align with the i18next documentation and also avoid calling init multiple times.

services/api/src/lib/i18n-adapter.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  getFixedT(
+    lng: string | readonly string[],
+    ns?: string | readonly string[],
+    keyPrefix?: string
+  ): TFunction {
+    return i18next.getFixedT(lng, ns, keyPrefix)
+  }
```
smonn, line 29: Use this to get a `t` function based on a user's locale, e.g.:

```ts
const t = this.i18n.getFixedT(user.locale, 'some-namespace')
```

services/api/src/lib/mailer-adapter.ts
```diff
@@ -0,0 +1,73 @@
...
+      emailFrom: string
+      emailName: string
+      emailTransport: 'smtp'
+      smtpHost: string
+      smtpPort: number
+      smtpUser: string
+      smtpPassword: string
+    }
+
+export default class MailerAdapter {
```
smonn, line 25: Replaces `SendgridAdapter`

IanPirro (approved):
Sweet 💯

services/api/package.json
```diff
@@ -24,14 +24,15 @@
     "@types/crypto-js": "4.0.2",
```
dallashuggins, line 24: When installing I got a new `package-lock.json` file

dallashuggins (commented):
Tried out both SendGrid and SMTP and getting errors for each. SMTP tells me I have an incorrect password when I know I don't, so still troubleshooting that but it could be the configuration; I have a hash in my password so I'm not sure if it's cutting off the rest of the characters and/or adding the quotes to the string when I add them. The bid notifications never get added and therefore don't get sent. The purchase notifications were getting added at least. But I have gotten both Unauthorized and Forbidden errors for SendGrid using the 2 accounts I have access to, but I noticed logging in to the one we created prior, it said I needed to contact support because there was an issue with my account, so that could be the source of the Unauthorized error.

dallashuggins (approved):
SMTP and SendGrid are successfully completing the notifications process. All of the strange behavior I was seeing seems to have been because of the configuration. When getting that down, it started going through. I only _received_ emails via SMTP, but received both bid and purchase emails with that. But SendGrid no longer throws errors when sending the API request so I can tell the configuration is right, but I haven't received those emails yet. I don't see any issue though, right up until the data is sent to SendGrid the console logs look right, so it seems like it has to be a delay with SendGrid. Especially if you got emails earlier. This is awesome! 🔥

smonn:
> SMTP and SendGrid are successfully completing the notifications process. All of the strange behavior I was seeing seems to have been because of the configuration. When getting that down, it started going through. I only _received_ emails via SMTP, but received both bid and purchase emails with that. But SendGrid no longer throws errors when sending the API request so I can tell the configuration is right, but I haven't received those emails yet. I don't see any issue though, right up until the data is sent to SendGrid the console logs look right, so it seems like it has to be a delay with SendGrid. Especially if you got emails earlier. This is awesome! 🔥

Had success with a new SendGrid account, so I have confirmed that the SendGrid setup works. Must be something odd with the old SendGrid account we've been using.

PR #117, opened 2021-11-29
Update configuration import

smonn:


services/api/src/configuration/index.ts
```diff
@@ -2,7 +2,7 @@ import { DEFAULT_CURRENCY } from '@algomart/schemas'
 import * as Currencies from '@dinero.js/currencies'
 import env from 'env-var'
 
-import { MailerAdapterOptions } from '@/lib/mailer-adapter'
+import { MailerAdapterOptions } from '../lib/mailer-adapter'
```
smonn, line 5: This import was breaking knex migrations. Seems like `tsconfig-paths` is not working as expected.

dallashuggins (approved):
🙏

PR #123, opened 2021-11-30
[#76] Task for checking wire transfers

### LINKED ISSUE #76: Scheduled task for checking wire payments

dallashuggins:
### Changes
- Scheduled task for updating wire payment method statuses
- Adding to the payment statuses task to support checking for wire transfers

Closes #76

### Testing
- In the web's env file, set `NEXT_PUBLIC_WIRE_PAYMENT_ENABLED` to `true`
- Create a purchase or auction type pack in the CMS with a price higher than 3000 (assuming USD is the currency used for web/api)
- Purchase pack and save wire details (or keep page open to reference)
- Mock wire transfer using [this endpoint](https://developers.circle.com/reference#payments-payments-mock-create); the `trackingRef` will match the `trackingRef` in the response (or Tracking Reference in the UI).

### Considerations
We are looking at payments within the past 24 hours, since we have to get all wire transfer payments and check for a matching source ID. Since an amount is not specified when creating a wire account using Circle's API and there's no payment record in Circle until its initiated by the user, the user _could_ send a wrong amount and this task would not be aware of this. Or perhaps if there was 2 wire payments done to correct the amount, it would not pick up on this either. 

Due to this, there should always be a way for a customer to contact customer support in the event something like this happened. Also, there at some point likely should be a way to aid customer support in verifying this. Though this wouldn't be straightforward, as the user could have made other wire payments using the same bank account, at least if we're planning to support reusing a bank account for a payment, which I'm assuming we will be. So maybe a manual option to award them the pack and a dashboard with all payments in Circle's API associated with that bank account, as well as a list of packs they've purchased. 

### To-do
Possible to-do to discuss: Not removing the claim on failure, but instead sending an email to the user and adding an expiration date to the payment + a task that looks for expired dates and removes a claim.

### DIFF
--- a/apps/cms/.env.example
+++ b/apps/cms/.env.example
@@ -5,8 +5,6 @@ PORT=8055
 PUBLIC_URL="http://localhost:8055"
 # Quieter logs (see: https://docs.directus.io/reference/environment-variables/#general )
 # LOG_LEVEL="warn"
-# Currency for application. Make sure this matches the api and web packages.
-CURRENCY=
 
 ####################################################################################################
 ## Database
--- a/apps/web/src/contexts/payment-context.tsx
+++ b/apps/web/src/contexts/payment-context.tsx
@@ -28,7 +28,7 @@ import collectibleService from '@/services/collectible-service'
 import { getExpirationDate } from '@/utils/date-time'
 import { encryptCardDetails } from '@/utils/encryption'
 import { toJSON } from '@/utils/form-to-json'
-import { formatFloatToInt, isGreaterThanOrEqual } from '@/utils/format-currency'
+import { formatFloatToInt, isGreaterThan } from '@/utils/format-currency'
 import { poll } from '@/utils/poll'
 import {
   maximumBidForCardPayments,
@@ -466,7 +466,7 @@ export function usePaymentProvider({
         if (
           !Environment.isWireEnabled ||
           (Environment.isWireEnabled &&
-            !isGreaterThanOrEqual(bid, maximumBidForCardPayments))
+            !isGreaterThan(bid, maximumBidForCardPayments))
         ) {
           setLoadingText(t('common:statuses.Authorizing card'))
 
--- a/services/api/src/modules/payments/payments.service.ts
+++ b/services/api/src/modules/payments/payments.service.ts
@@ -30,8 +30,9 @@ import { PaymentCardModel } from '@/models/payment-card.model'
 import { UserAccountModel } from '@/models/user-account.model'
 import NotificationService from '@/modules/notifications/notifications.service'
 import PacksService from '@/modules/packs/packs.service'
-import { formatIntToFloat } from '@/utils/format-currency'
+import { formatFloatToInt, formatIntToFloat } from '@/utils/format-currency'
 import {
+  convertFromUSD,
   convertToUSD,
   currency,
   isGreaterThanOrEqual,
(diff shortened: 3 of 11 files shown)


apps/cms/.env.example
```diff
@@ -5,8 +5,6 @@ PORT=8055
 PUBLIC_URL="http://localhost:8055"
 # Quieter logs (see: https://docs.directus.io/reference/environment-variables/#general )
 # LOG_LEVEL="warn"
-# Currency for application. Make sure this matches the api and web packages.
-CURRENCY=
```
dallashuggins, line 9: This is specified within the application itself now, so removing.

services/api/src/modules/payments/payments.service.ts
```diff
@@ -459,23 +462,73 @@ export default class PaymentsService {
...
+    if (sourcePayment.status === PaymentStatus.Failed) {
+      await this.packs.claimPack(
+        {
+          packId: payment.packId,
+          claimedById: null,
+          claimedAt: null,
+        },
+        trx
+      )
+    }
```
dallashuggins, line 523: Should we remove claim if payment fails? Or just mark as a failure and possibly email the client?

apps/web/src/contexts/payment-context.tsx
```diff
@@ -466,7 +466,7 @@ export function usePaymentProvider({
         if (
           !Environment.isWireEnabled ||
           (Environment.isWireEnabled &&
-            !isGreaterThanOrEqual(bid, maximumBidForCardPayments))
+            !isGreaterThan(bid, maximumBidForCardPayments))
```
dallashuggins, line 469: Noticed an error where if you make a bid over $2999 it wouldn't go through, since the card form is still displayed. This allows a $3000 bid, and then all future bids the bank account form will display. The bid needs to be placed though, it doesn't change the form depending on the bid amount, so this should be updated.

dallashuggins: This is just a temporary solution to allow an exact bid of 3000. We need the form displayed in the checkout template to change depending on what bid is placed. If the bid is over 3k, unless there was already a bid for 3k and therefore the current bid has hit that amount, currently the form will still submit as a credit card payment flow, but fail since the correct fields weren't submitted. I added more details [here](https://github.com/deptagency/algomart/issues/125). I started trying to implement, and could not get the checkout template to update. I figured this may be worth further discussion since I'm not completely sure how it should behave, but would appreciate help on how to accomplish at that point.

services/api/src/modules/payments/payments.service.ts
IanPirro: No strong opinion, but curious what others think. Seems reasonable to email the customer and maybe give them X time to resubmit (maybe 48hrs etc.) before removing the claim info on the pack?

dallashuggins: Sweet, I like that idea. We could leverage the expiration field already in the database, and have a task that checked for a past expiration date which removed the claim, so that there's only one chance to resubmit. It may need to be longer than 48 since wires take so long. Will have to research and see if I can find how long would be reasonable.

dallashuggins: Nevermind on the expiration, was thinking of bids. But we could add an expiration field to payments.

PR #129, opened 2021-12-02
3D Secure support

smonn:
**Note: Waiting to hear back from Circle on how to test the 3DS not supported flow. Otherwise, this PR is ready for review.** 

### Changes
- 3DS support to enable payments from other countries. 
  - Always attempts 3DS flow initially
  - Verification using cvv will be carried out if the initial payment fails

#### 3DS - Challenge Flow
With this flow, the user will be redirected to a sample bank login page provided by Circle. Circle provides the test password that you can use to authenticate. Afterwards, you're redirected to the success page. 

1. Add pack with an amount of $5.66
2. Once pack is generated, go through card payment flow

#### 3DS - Frictionless flow with success
The user will be redirected and immediately returned back to the application's success page. This would happen if the user had already recently authenticated with their bank.

1. Add pack with an amount of $5.69
2. Once pack is generated, go through card payment flow

#### 3DS - Frictionless flow with failure
The user will be redirected and immediately returned back to the application's failure page, to show the behavior if the verification failed. If the payment fails, a cvv payment will be attempted. 

1. Add pack with an amount of $5.72
2. Once pack is generated, go through card payment flow


apps/api/src/modules/payments/payments.service.ts
```diff
@@ -423,16 +426,33 @@ export default class PaymentsService {
...
+        metadata: {
+          ...metadata,
+          sessionId: SHA256(user.id).toString(enc.Base64),
+        },
         amount: {
           amount: priceInUSD,
           currency: DEFAULT_CURRENCY,
         },
-        verification: verification,
+        verification: CirclePaymentVerificationOptions.three_d_secure,
```
smonn, line 446: Verify this line is accurate

smonn: Add second attempt using CVV for verification when failing to create payment with 3DS.

apps/api/src/lib/circle-adapter.ts
```diff
@@ -98,31 +94,12 @@ function toBankAccountBase(
...
-  } else {
-    finalStatus = PaymentCardStatus.Pending
-  }
-  return finalStatus
+function toCardStatus(status: CircleCardStatus): PaymentCardStatus {
+  return {
+    [CircleCardStatus.Complete]: PaymentCardStatus.Complete,
+    [CircleCardStatus.Failed]: PaymentCardStatus.Failed,
+    [CircleCardStatus.Pending]: PaymentCardStatus.Pending,
+  }[status]
```
dallashuggins, line 102: We don't need to have these conditionals any longer, because Circle will resolve itself based on the cvv, 3DS, and avs checks already.

apps/api/src/modules/payments/payments.service.ts
```diff
@@ -541,25 +544,46 @@ export default class PaymentsService {
     // Attempt to find card (cardId could be source or db ID)
     const card = await PaymentCardModel.query(trx).findById(cardId)
 
-    // Create payment using Circle API
+    // Circle only accepts loopback addresses
+    const verificationHostname = Configuration.webUrl.includes('localhost')
+      ? 'http://127.0.0.1:3000'
+      : Configuration.webUrl
```
dallashuggins, line 550: Let me know if you have any ideas how to setup better. Loopback addresses can be used for the Circle verification URLs, but localhost cannot.

apps/api/src/modules/payments/payments.service.ts
```diff
@@ -541,43 +544,75 @@ export default class PaymentsService {
...
+    let payment: ToPaymentBase | undefined
+    if (!paymentResponse) {
+      // Create cvv payment
+      const cvvPayment = await this.circle
+        .createPayment({
+          ...basePayment,
+          ...encryptedDetails,
+          verification: CirclePaymentVerificationOptions.cvv,
+        })
+        .catch(() => null)
```
dallashuggins, line 596: Need to verify Circle will immediately fail, and not fail on the verification failure URL page, so that we can run cvv verification after. Through testing I was only finding that errors resulted in the failure URL being hit, but talking with Circle initially, we expected to get an immediate failure so we can run secondary verification, so waiting to hear back. 🤞

jakerainis (approved):

PR #130, opened 2021-12-02
CMS currency updates

smonn:
Also enables localization support and upgrades Directus to 9.1.2.

Kapture 2021-12-02 at 17 15 22


IanPirro (approved):
Looks 💯 , this field has been driving me crazy 🙌

dallashuggins (approved):
Nice! 🙌 Not sure why the field started behaving like that, I swear that was one of the original fixes. But also have been wanting to get this working better. Thanks for tackling that!

PR #134, opened 2021-12-06
Upgrade Directus to v9.2.0

smonn:
This version seems to work well on my end and I don't see the issues from before.


IanPirro (approved):
Pulled it down to try out, didn't get any of the call stack issues either :shipit:

dallashuggins (approved):
Not seeing those errors locally 🙌

PR #151, opened 2021-12-09
Initial support for Algorand Wallet

smonn:
- initial setup for algorand wallet
- allow any in rest args


apps/web/src/libs/algorand-adapter.ts
```diff
@@ -0,0 +1,125 @@
...
+  Algodv2,
+  makePaymentTxnWithSuggestedParamsFromObject,
+  Transaction,
+} from 'algosdk'
+
+import { EventEmitter } from './event-emitter'
+
+const algosdkLoader = import('algosdk')
+
+export enum ChainType {
```
jakerainis, line 11: Do we want to put some of these interfaces in schemas? Perhaps there could be some cross over with the AlgorandAdapter in the API?

apps/web/src/libs/event-emitter.ts
```diff
@@ -0,0 +1,35 @@
+export interface Subscriber {
+  (...args: any[]): void
+}
+
+export class EventEmitter {
+  private _subscribers: { [event: string]: Subscriber[] } = {}
+
+  public subscribe(event: string, function_: Subscriber) {
```
jakerainis, line 8: `function_` is a little ambiguous, unless it means something I don't understand, wouldn't this just be something like `callback`?

apps/web/src/libs/event-emitter.ts
```diff
@@ -0,0 +1,35 @@
...
+      this._subscribers[event] = this._subscribers[event].filter(
+        (subscriber) => subscriber !== function_
+      )
+    }
+  }
+
+  public emit(event: string, ...args: unknown[]) {
+    if (this._subscribers[event]) {
+      for (const subscriber of this._subscribers[event]) {
+        setTimeout(() => {
```
jakerainis, line 29: Curious why we need a timeout. Is it just to bump it to the bottom of the callstack?

jakerainis (approved):
Left some minor housekeeping questions, but feel free to ignore as I know your intention was to just add bare minimum so you and Dallas could work concurrently.

apps/web/src/libs/event-emitter.ts
smonn: Good point! I'll rename it to `subscriber`.

apps/web/src/libs/event-emitter.ts
smonn: In Node.js this would've been `setImmediate`, but that's not an official JS API. There is a polyfill but didn't wanna add that just for this. But yes, the idea is to push it to the bottom of the callstack.

apps/web/src/libs/algorand-adapter.ts
smonn: Potentially. For now let's keep them separate as they have different priorities.

dallashuggins (approved):
Thanks for adding! 🙌 Was able to try out connecting to my wallet.

PR #137, opened 2021-12-07
[#131] Payment flow

dallashuggins:
### Changes
- Combining the purchase form components (i.e., purchase nft form, bank account form) into one purchase form component, but separating the functionality by the payment method selected. The card-details component now also lives in purchase form in the shared folder. 
- `Breadcrumbs` component for displaying the navigational breadcrumbs (not traditional breadcrumbs with href, but a function that modifies state)
- `Cards` component for displaying the payment methods.
- New checkout flow (see below)

#### Checkout flow
1. Select a method
2. Relevant form displayed (cards or wire)
3. Summary (method, pack name, price)
4. Form is submitted
5. Success/error screen appears or you're kicked back to the form with any identified validation errors

**Closes #131** 

#### Payment methods
Screen Shot 2021-12-07 at 1 53 02 PM

#### Methods with cc disabled
Screen Shot 2021-12-07 at 1 51 54 PM

#### Payment Information - pack for purchase with credit card method
Screen Shot 2021-12-07 at 1 53 11 PM

#### Payment summary - pack for purchase with credit card method
Screen Shot 2021-12-07 at 1 53 20 PM

#### Payment Information - pack for auction, bid using credit card method
Screen Shot 2021-12-07 at 3 24 14 PM

#### Payment Information - pack for auction, bid using credit card method
Screen Shot 2021-12-07 at 3 24 28 PM

#### Payment Information - placing bid using wire payment method
Screen Shot 2021-12-07 at 1 52 04 PM

#### Payment summary - wire transfer method summary for bid
Screen Shot 2021-12-07 at 1 52 29 PM

#### Payment - wire transfer method form
Screen Shot 2021-12-07 at 3 28 45 PM
Screen Shot 2021-12-07 at 3 29 08 PM

#### Payment - wire transfer method summary
Screen Shot 2021-12-07 at 3 32 45 PM

#### Payment - wire transfer method confirmation of bank account creation
Screen Shot 2021-12-07 at 3 33 02 PM

Issue #104, opened 2021-11-19
Sending notification to the user when a transfer is complete


dallashuggins:
## Overview

We need to send a user an email notification that the transfer was complete upon successful transfer to the Circle wallet address.

dallashuggins:
At this time, we can send a notification to a user when the transfer is considered paid, letting them know it's available for transfer. However, since we are minting collectibles which are owned, they will be notified in their account sooner than this to tell them their transfer is ready, but will receive the email once paid. If we add a claimed by column for packs, then we could consider them an owner only after their payment has settled (see #169).

dallashuggins:
Current functionality in #156 - Initially I added the following snippet to the `updatePaymentStatuses` method where the status is checked if it's `Paid`. I took this out, because users will still be able to transfer the assets before the payment has been settled, so it could add confusion. If the payment status was `paid` during the time of payment (common with transfers of crypto), the user wouldn't receive an email (unless the status came back `pending`). For cards, payments will have a status of `confirmed` for ~3 days, but the user will be able to transfer the pack still, but will receive an email when paid. This is confusing to the user since they already transferred the asset. This could be useful for wires however. I added an email for `paymentSuccess` but keeping out sending the notification until we're utilizing claimed by. 

```
        // Create purchase success notification to be sent to user
        const packWithBase = await this.packs.getPackById(payment.packId)
        if (packWithBase) {
          await this.notifications.createNotification(
            {
              type: NotificationType.PaymentSuccess,
              userAccountId: payment.payerId,
              variables: {
                packTitle: packWithBase.title,
              },
            },
            trx
          )
        }
```

PR #156, opened 2021-12-10
[#102, #142, #103, #105, #161] Wallet transfers

### LINKED ISSUE #102: API support for wallet transfers
## Overview
We need to support creating a wallet and getting transfers (querying by wallet ID) in order to provide the user with a wallet address to send to, and to verify that payment went through. We also need a table in the database for the wallet address they provide, so we can check for payments from that address. In addition, we should have an association between the payment table and the wallet table, so we can create a payment eventually, and then have this record in their profile. 

_Endpoints_
- Create blockchain address ([endpoint](https://developers.circle.com/reference#payments-on-chain-addresses-create))
  - Must use the wallet ID for the merchant.
- Get wallets ([endpoint](https://developers.circle.com/reference#payments-on-chain-wallets-get))
  - Find the "merchant" type wallet
  - This doesn't need to be it's own endpoint, but needs to be used in the request to create a blockchain address.
- Get transfers ([endpoint](https://developers.circle.com/reference#payouts-transfers-get))
  - I wasn't able to test this, I _think_ because the address generated is on MainNet, even in the sandbox. I wrote Circle to let them know. The blockchain address they return in the sandbox is an address that's been around for over a year, longer than we've had this Circle account.

### LINKED ISSUE #103: Front-end support for wallet transfers
## Overview

In addition to the API endpoints supporting the creation of a blockchain address, we need to hook up with WalletConnect SDK, where the user connects their wallet and then we send a transaction for them to sign, and finally award the pack. 

## Additional context

_Steps_
- Hook up API endpoints for creating a blockchain address, etc.
- WalletConnect => connect and send transactions
- Query for the payment
  - Do we need to monitor for settlement? Or is this immediate after sending a transaction?

### LINKED ISSUE #105: Task for checking for successful wallet transfers
## Overview

When we have the wallet address from the user, we can start polling for the status of that wallet address. 

When found, we should add a payment record in the database associated with the wallet. 

## Additional context

We need to confirm the price of the wallet transfer as well. #102 needs to be complete.

### LINKED ISSUE #142: Implement support for Algorand Wallet via WalletConnect

### LINKED ISSUE #161: Crypto payments documentation
## Overview

There is a place for a hyperlink in the designs for the crypto flow. It states "We'll walk your through what is needed" and _what is needed_ is a hyperlink. We need to create a page that has details on how to create a wallet, etc. so we can link it here. 

## Additional context

Screen Shot 2021-12-17 at 8 47 36 AM

dallashuggins:
### Changes
- Built on top of #137 
- Payment method addition - Pay with Crypto Wallet
  - Options
    - Copy the address for manually sending USDC
    - _OR_ use Wallet Connect to initiate USDC payment
  - Pack is awarded when a transfer is found for the purchase. 
- Circle API functionality: create a blockchain address, get transfers, get merchant wallet. API endpoint for creating a blockchain address, getting a transfer, and creating a payment when a transfer is found. 
- Components for use when the `crypto` method is selected from the payment methods. 
- Change of routes => `/checkout?pack=[packSlug]` is now `checkout/[packSlug]` with an additional page for the method chosen (i.e., card, crypto, wire) => `/checkout/[packSlug]/[method]`. The `step` parameter is used to change between the form page (`?step=details`) and the summary page (`?step=summary`). 
- New wallet instructions modal triggered by the "what is needed" link in the crypto payment flow. 

Closes #102 
Closes #142 
Closes #103 
Closes #105 
Closes #161 
Progress on #104 

#### Algorand Wallet
1. Go to the Algorand Wallet app and scan QR code to connect
2. In app, click to purchase using Crypto Wallet
3. Click to "Connect" using Algorand Wallet
4. After connecting (should see wallet ID listed), click Purchase
5. Open Algorand Wallet and accept transaction
6. You should see a "Success" screen afterwards

#### Manually (using Algorand Wallet)
1. In app, click to purchase using Crypto Wallet and copy the address
2. Go to the Algorand Wallet app
3. Submit a transaction for USDC
4. Click to Check for Payment until you see a "Success" screen (or error)

#### 1) Crypto purchase flow screen video
https://user-images.githubusercontent.com/13154207/146572836-7048345c-0c00-483e-8207-1c86716a3022.mov

#### 2) Crypto purchase flow screen video - mobile portion of flow
https://user-images.githubusercontent.com/13154207/146572905-216a6a6a-afa5-4dbc-b240-5b1724ec7671.mp4

#### 3) Wire purchase flow screen video
https://user-images.githubusercontent.com/13154207/146572987-7902188f-8e3c-4cee-b2bd-455aca00e1b4.mov

#### 4) Card purchase flow screen video
https://user-images.githubusercontent.com/13154207/146572998-e69842fc-5283-4d47-bef3-24e482b016e6.mov

### Situations to handle better
- Confirming payment amount before payment creation for crypto flow: See `@TODO` note inside `createTransferPayment` method in `payments.service.ts`. Currently we'll error if the amount is incorrect, but the payment was valid, so this will require customer support intervention.


apps/web/src/components/bid-activity/bid-activity.tsx
```diff
@@ -82,10 +78,10 @@ export default function BidActivity({
           const avatar = avatars[bid.externalId]
           const createdAtDateTime = new Date(bid.createdAt)
           const meetsReservePrice =
-            reservePrice && isGreaterThanOrEqual(bid.amount, reservePrice)
+            !!reservePrice && isGreaterThanOrEqual(bid.amount, reservePrice)
```
dallashuggins, line 81: When the reserve price wasn't included in the pack template, "0" would display in the activity table. This fixes that.

services/api/src/lib/circle-adapter.ts
```diff
@@ -107,40 +112,45 @@ function toCardBase(response: CircleCard): ToPaymentCardBase {
...
+      finalStatus = PaymentStatus.Failed
+      break
+    case CircleTransferStatus.Complete:
+      finalStatus = PaymentStatus.Paid
+      break
+    case CircleTransferStatus.Pending:
+      finalStatus = PaymentStatus.Pending
+      break
+    default:
+      finalStatus = PaymentStatus.Pending
```
dallashuggins, line 141: Including transfer statuses as payment statuses, since we need the same info for creating a payment in the database, and transfer has a very similar response body.

services/api/src/lib/circle-adapter.ts
```diff
@@ -286,6 +334,63 @@ export default class CircleAdapter {
...
+    const response = await this.http
+      .get('v1/transfers', {
+        searchParams,
+      })
+      .json<CircleResponse<CircleTransfer[]>>()
+
+    if (isCircleSuccessResponse(response)) {
+      const transfer = response.data.find(
+        (transfer) => transfer.destination.address === destinationAddressId
+      )
```
dallashuggins, line 367: Since we're generating a new address for every transaction, there should only be one transfer for that address, though there's no way to search by the address in Circle.

services/api/src/modules/payments/payments.service.ts
```diff
@@ -657,6 +745,22 @@ export default class PaymentsService {
...
+            transfer &&
+            (payment.status !== transfer.status || !payment.transferId)
+          ) {
+            await PaymentModel.query(trx).patchAndFetchById(payment.id, {
+              status: transfer.status,
+              transferId: transfer.externalId,
+            })
+            updatedPayments++
+          }
+        }
```
dallashuggins, line 763: Handles updating payment statuses for crypto payments.

afraser (approved):

apps/web/src/pages/checkout/[packSlug].tsx
```diff
@@ -0,0 +1,184 @@
...
+import { urls } from '@/utils/urls'
+
+export interface CheckoutPageProps {
+  address?: string
+  auctionPackId: string | null
+  currentBid: number | null
+  release: PublishedPack
+}
+
+export default function Checkout({
```
Comment (dallashuggins) [inline apps/web/src/pages/checkout/[packSlug].tsx:33]: Changed the route for the checkout page to be `/checkout/[packSlug]` as opposed to `/checkout?pack=[packSlug]`, since pack is a required field and now that we're using query params for method/step, it reduces complexity to not have to keep adding `pack` back onto the route.

apps/web/src/environment.ts
```diff
@@ -69,4 +69,9 @@ export const Environment = {
     const isEnabled = this.config('NEXT_PUBLIC_WIRE_PAYMENT_ENABLED', '')
     return isEnabled.toLowerCase() === 'true'
   },
+
+  get isCryptoEnabled() {
+    const isEnabled = this.config('NEXT_PUBLIC_CRYPTO_PAYMENT_ENABLED', '')
+    return isEnabled.toLowerCase() === 'true'
+  },
```
dallashuggins, line 76: New feature flag for enabling the payment flow for cryptocurrencies

apps/web/src/pages/checkout/[packSlug].tsx
```diff
@@ -0,0 +1,184 @@
...
+  useEffect(() => {
+    if (
+      query.method &&
+      Object.values(CheckoutMethod).includes(query.method as CheckoutMethod)
+    ) {
+      setMethod(query.method as CheckoutMethod)
+    } else {
+      setMethod(null)
+    }
+  }, [query.method, setMethod])
```
Comment (dallashuggins) [inline apps/web/src/pages/checkout/[packSlug].tsx:65]: This changes the method based on the query param, so we can send the user to `/checkout/[packSlug]?method=card` directly, and it will direct them to the card payment flow.

apps/web/src/pages/checkout/[packSlug].tsx
```diff
@@ -0,0 +1,184 @@
...
+          permanent: false,
+        },
+      }
+    }
+  }
+
+  // Get release based on search query
+  const address = await ApiClient.instance.createWalletAddress({
+    idempotencyKey: uuid(),
+  })
```
Comment (dallashuggins) [inline apps/web/src/pages/checkout/[packSlug].tsx:172]: Since we cannot use `serverSideProps` with a component, we have to retrieve the wallet address here. If no address is created and therefore set, the crypto option will be disabled in the methods form, and it will redirect to the methods form if the route is hit directly when no address is available.

apps/web/src/utils/format-currency.ts
```diff
@@ -20,7 +21,7 @@ export function formatCurrency(
   if (amount === null || amount === undefined) {
     amount = 0
   } else if (typeof amount === 'string') {
-    amount = Number.parseFloat(amount) * 100
+    amount = Math.round(Number.parseFloat(amount) * 100)
```
dallashuggins, line 24: The function was failing for certain numbers. I can't remember if this one of the problem #s, but if for instance `10.13` was submitted, it would get an amount of `10.1299999` instead, and Dinero would say the number was invalid. This fixes that issue.

apps/web/src/pages/checkout/[packSlug]/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+  if (!Environment.isWireEnabled && !Environment.isCryptoEnabled) {
+    return {
+      redirect: {
+        destination: urls.checkoutPackWithMethod
+          .replace(':packSlug', context?.params?.packSlug as string)
+          .replace(':method', 'card'),
+        permanent: false,
+      },
+    }
+  }
```
Comment (dallashuggins) [inline apps/web/src/pages/checkout/[packSlug]/index.tsx:80]: Instead of displaying the methods screen, if the feature flags aren't enabled, direct the user to the `/card` endpoint where they can fill out cc info.

apps/web/src/pages/checkout/[packSlug]/[method].tsx
```diff
@@ -0,0 +1,215 @@
...
+  ) {
+    return {
+      redirect: {
+        destination: urls.checkoutPackWithMethod
+          .replace(':packSlug', context?.params?.packSlug as string)
+          .replace(':method', 'card'),
+        permanent: false,
+      },
+    }
+  }
```
Comment (dallashuggins) [inline apps/web/src/pages/checkout/[packSlug]/[method].tsx:126]: Redirects if a) the checkout method isn't recognized and b) if another method is accessed when feature flags aren't enabled.

apps/web/src/contexts/payment-context.tsx
```diff
@@ -108,6 +131,29 @@ export function usePaymentProvider({
...
+      setStatus(status)
+      const step = status === CheckoutStatus.form ? 'details' : 'summary'
+      const path = `${asPath.split('?')[0]}?step=${step}`
+      if (path !== asPath) {
+        return push(`${asPath.split('?')[0]}?step=${step}`)
+      }
+      return
+    },
+    [asPath, push]
+  )
```
dallashuggins, line 150: Change both the status and the path to reflect accurately, which can be used in order to return to the form page after an error is found post-submission.

apps/web/src/components/purchase-form/crypto/crypto-purchase-form.tsx
```diff
@@ -0,0 +1,145 @@
...
+          destinationAddress: address,
+        })
+        .catch(() => null)
+      if (!transferPayment) {
+        // While this shouldn't happen, there's a possibility the payment may still have worked
+        // @TODO: Find way to handle this better - possibly send to customer support email or direct to contact
+        setError(t('forms:errors.paymentNotCreated'))
+        setStatus(CheckoutStatus.error)
+        return
+      }
```
dallashuggins, line 75: If the user sends the wrong amount, this condition will be hit.

apps/web/src/libs/wallet-connect-adapter.ts
```diff
@@ -86,24 +86,34 @@ export class WalletConnectAdapter extends EventEmitter implements IConnector {
   }
 
   public async signTransaction(transaction: Transaction, message?: string) {
-    if (!this._connector) throw new Error('WalletConnect not initialized')
+    try {
```
dallashuggins, line 89: The try/catch allows catching a submitted transaction that was declined. Now the error screen will appear in this case. The user can select to Try Again.

apps/web/src/components/purchase-form/cards/card-form.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const [promptLeaving, setPromptLeaving] = useState(false)
+  const isAuctionActive =
+    release?.type === PackType.Auction &&
+    isAfterNow(new Date(release.auctionUntil as string))
+  useWarningOnExit(promptLeaving, t('common:statuses.processingPayment'))
+
+  const handleSubmitPurchase = useCallback(
+    async (event: FormEvent<HTMLFormElement>) => {
+      event.preventDefault()
+      setPromptLeaving(true)
```
dallashuggins, line 44: Setting `setPromptLeaving` causes the app to throw a message confirming changing routes when there's a form error. This is because now we'll be changing the URL param based on if the user is viewing the form details or the summary, so it sends the user back to the `?step=details` page from `?step=summary` and throws the prompt. If they click "no" they won't be able to click "Go to Summary" since it the route won't be able to change back to summary (since it never changed back to details). Should we remove the prompt?

smonn (approved):

PR #157, opened 2021-12-10
[#149] Added AlgoExplorer adapter and address page

### LINKED ISSUE #149: Display collectibles based on a provided Algorand account address

jakerainis:
## Overview
This PR does not actually touch the showcase functionality at all. It was originally going to replace it, but at the moment, we're still planning to use the showcase with its original functionality for existing Firebase users. Instead, this PR adds a new page and API integration to display collectibles associated with a given account address:

1. User (anonymous or auth'd) accesses page at `/address/:algoAddress`
2. That param is passed along to the API, which has a new `AlgoExplorerAdapter`
3. The adapter looks up the account by the provided address and returns account information, which includes an array of ASAs and their associated asset-id.
4. The API queries our database for collectibles with those ids, matches them to a template, and returns collectible information to display on the page, which is sorted or paginated based on other query string arguments optionally provided in the initial request.

Additional notes:
- IF an address is invalid, malformed, or causes AlgoExplorer to not return an account, text is displayed indicating the address does not exist.
- IF an address is valid, but does not contain assets within our DB, text is displayed indicating that there are no collectibles owned by the provided address.
- SINCE the collectible viewer might not display an avatar/username, I've augmented the viewer to not show an avatar, and instead of a username, display a formatted address (e.g. `AAAAAA...AAAAAA`)
- SINCE interfacing with AlgoExplorer takes place on `mainnet`|`testnet`|`betanet`, I've added a new environment variable `ALGOD_ENV` that takes one of those three values. 

Closes #149 

## Screenshots
Desktop:
screencapture-localhost-3000-address-KI7GYT2Y5PXYNLKQD562WTJDV2CVA6YRNHEFH5LA6OIVISSYAQR6TC3LE4-2021-12-10-16_03_37

Mobile (note truncated address):
Screen Shot 2021-12-10 at 4 04 37 PM

Address is valid but does not contain any collectibles:
screencapture-localhost-3000-address-EX3L3OW2EOCHDJCUX3XDGDIQQZFVNOSSRVCG4ZK73VLZNZ2EO4ZBT2HE6U-2021-12-10-16_03_53

Address not found:
screencapture-localhost-3000-address-EX3L3OW2EOCHDJCUX3XDGDIQQZFVNOSSRVCG4ZK73VLZNZ2EO4ZBT2HE6s-2021-12-10-16_04_10


apps/web/src/components/profile/my-profile-wallet.tsx
```diff
@@ -8,16 +8,14 @@ import css from './my-profile-wallet.module.css'
...
 export default function MyWallet() {
   const { user } = useAuth()
   const { t } = useTranslation()
 
   const formattedAddress = useMemo(() => {
-    const addr = user?.address as string
-    const first6 = addr.slice(0, 6)
-    const last6 = addr.slice(-6, addr.length)
-    return `${first6}...${last6}`
+    return formatAlgoAddress(user?.address || '')
```
jakerainis, line 18: Moved to utils file since it's being used in multiple places.

apps/web/src/templates/algo-address-template.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const Title = () => {
+    return (
+      <Heading className={css.pageTitle} level={1}>
+        {t('collection:algoAddressPage.Algorand Account')}:
+        <div>
+          {algoAddressIsInvalid ? (
+            <Address />
+          ) : (
+            <ExternalLink
+              href={`https://algoexplorer.io/address/${algoAddress}`}
```
jakerainis, line 67: This is just an external link to AlgoExplorer, so I've hardcoded it to mainnet, which I think is fine as it won't impact any functionality _within_ our app. But just wanted to point it out.

services/api/src/configuration/index.ts
```diff
@@ -54,6 +54,14 @@ export const Configuration = {
     return env.get('ALGOD_PORT').default('4001').asPortNumber()
   },
 
+  get algodEnv() {
+    return env
+      .get('ALGOD_ENV')
+      .required()
+      .default('testnet')
+      .asEnum(['betanet', 'testnet', 'mainnet'])
+  },
```
jakerainis, line 63: As defined in `.env.example`, this value can be used to determine which flavor of the blockchain we want to interface with. In this PR, this determines which API endpoint to hit on AlgoExplorer, but could also see this being used for the wallet SDKs? 

Is the name `ALGOD_ENV` okay?

afraser (approved):

services/api/src/lib/algoexplorer-adapter.ts
```diff
@@ -0,0 +1,54 @@
...
+interface AlgoExplorerAsset {
+  amount: number
+  'asset-id': number
+  creator: string
+  'is-frozen': boolean
+}
+
+interface AlgoExplorerAccount {
+  address: string
+  amount: number
```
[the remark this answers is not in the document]
jakerainis: Good question. We don't store or use anything related to this value in the application. I (partially) typed interface this based on the response we get from AlgoExplorer. To avoid confusion, maybe it's best to only type the parts we _do_ use. I'll make that change unless there are any objections.

services/api/src/lib/algoexplorer-adapter.ts
```diff
@@ -0,0 +1,54 @@
...
+import { Configuration } from '@/configuration'
+import { logger } from '@/utils/logger'
+
+/**
+ * AlgoExplorer APIv2 Docs:
+ * https://testnet.algoexplorer.io/api-dev/v2
+ */
+
+interface AlgoExplorerAsset {
+  amount: number
```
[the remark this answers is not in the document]
jakerainis: It's actually referred to as the AssetHolding (in the models here: [https://testnet.algoexplorer.io/api-dev/v2#model-AssetHolding](https://testnet.algoexplorer.io/api-dev/v2#model-AssetHolding). I can update the name to reflect that. 

Like I mentioned above, this is typed based on what comes back from their API. The only real value we need from it is the `asset-id`, so I can strip this interface down to only include what we need.

apps/web/src/templates/algo-address-template.tsx
jakerainis: You're correct, it wouldn't work outside of production. This was half laziness, but also half not wanting to add another (duplicative) environment variable. The `configureAlgoExplorerUrl` lives within the API, not the web front end. I could absolutely do that, it just seems like a lot of overhead to have an external link resolve correctly in a non-production environment. I'll defer to the team's concensus.

dallashuggins (approved):
Looks great! No comments other than what's already been mentioned. I was able to see the different states for the new address page. I didn't see the collectibles, since locally I use the Docker Algorand node without deploying to Testnet, but they look great in the screenshot. 🔥

apps/web/src/templates/algo-address-template.tsx
smonn: Could hardcode values based on the `ALGOD_ENV` value, but this is probably fine either way. There's a similar thing on the user's settings page.

smonn (approved):

PR #158, opened 2021-12-14
[MISC] UI & Docs Improvements

jakerainis:
A couple of things I caught when upkeeping another implementation. 

- In square depictions of a pack or a collectible, the image is now set to have a object-fit of `cover`. This prevents odd image ratios from displaying undesirably
- Updated some README information to reflect recent setup changes
- Linting & alphabetization


apps/web/src/contexts/payment-context.tsx
```diff
@@ -271,7 +271,6 @@ export function usePaymentProvider({
           .catch(async (error) => {
             const response = await error.response.json()
             mapCircleErrors(response.code)
-            setFormErrors({})
```
jakerainis, line 274: This line and the one on :378 was overriding the errors that are set in `mapCircleErrors` (called on previous line). As a result, if you entered a bad CC number or full name, the form would be in an error state, but would not display the validation errors.

terraform/README.md
```diff
@@ -78,7 +78,7 @@ and you are free to override any variables that do.
 | `email_transport`                 | Either `sendgrid` or `smtp`                                                                                                 |
 | `email_name`                      | The sender name                                                                                                             |
 | `smtp_host`                       | The SMTP hostname, only required when `email_transport` is set to `smtp`                                                    |
-| `smtp_post`                       | The SMTP port, only required when `email_transport` is set to `smtp`                                                        |
+| `smtp_post`                       | The SMTP port, a valid port is required when `email_transport` is set to `smtp`, otherwise, set to a non-falsy integer      |
```
IanPirro, line 81:
```suggestion
| `smtp_port`                       | The SMTP port, a valid port is required when `email_transport` is set to `smtp`, otherwise, set to a non-falsy integer      |
```

IanPirro (approved):

smonn (approved):
Nice, thanks for fixing these!

dallashuggins (approved):

PR #181, opened 2022-01-06
update readme

larpollack:
Quick PR to specify required node version in readme after encountering an error

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ This software is in a pre-release state. This means while we strive to keep it s
 
 ## ✅ Requirements
 
-- Node.js v16, npm v7 (manage version via [nvm][nvm])
+- Node.js v16.10 or greater (lts is v16.13.1 as of Jan 2022 and works well), npm v7 or greater (manage version via [nvm][nvm])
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -45,7 +45,7 @@ This software is in a pre-release state. This means while we strive to keep it s
 
 ## ✅ Requirements
 
-- Node.js v16, npm v7 (manage version via [nvm][nvm])
+- Node.js v16.10 or greater (lts is v16.13.1 as of Jan 2022 and works well), npm v7 or greater (manage version via [nvm][nvm])
```
jakerainis, line 48: Are we sure npm > 7 (e.g. v8) works? I haven't tried it myself, so I'm not sure. It strikes me that it should work but these things are nuanced within the monorepo.

larpollack: It's working for me at the moment, but definitely a good question

afraser (approved):

README.md
smonn: Should work with npm v8. As far as I know, the main difference is that npm v8 does not support Node v10.

smonn (approved):

PR #201, opened 2022-01-14
themes

afraser:


afraser:
I dumbed this down big-time since I last shared it. Themes are no longer defined in ts, they are simply scoped variables that are referenced by our tailwind theme. Custom dark mode styles can still be applied via the `dark:` syntax. We get a lot for free (in terms of no added css, by using the variables though).

apps/web/src/contexts/theme-context.tsx
```diff
@@ -0,0 +1,91 @@
...
+interface IThemeContext {
+  theme?: string
+  toggleTheme: () => void
+  setTheme: (theme?: string) => void
+}
+
+const ThemeContext = createContext<IThemeContext>({
+  theme: DEFAULT_THEME,
+  setTheme: () => {}, // eslint-disable-line @typescript-eslint/no-empty-function
+  toggleTheme: () => {}, // eslint-disable-line @typescript-eslint/no-empty-function
```
dallashuggins, line 41: Would it make sense to return null or add a comment inside to avoid the exception? Just an idea, this works too.

apps/web/src/styles/globals.css
```diff
@@ -7,9 +7,55 @@
...
+    --textDisabled: 102, 102, 102;
+
+    --bg: 236, 236, 236;
+    --bgCard: 255, 255, 255;
+    --bgPanel: 255, 255, 255;
+
+    --actionPrimary: 10, 17, 29;
+    --actionContrastText: 255, 255, 255;
+    --actionSecondary: 18, 220, 197; /* was "teal" */
+    --actionAccent: 2, 251, 194; /* was "green" */
```
dallashuggins, line 27: What is meant by this was teal? It still is, right?

dallashuggins: It may be helpful to document the colors next to each

dallashuggins (commented):
Awesome! 🔥 I like the approach, it's very clean and organized now. It could be helpful to add documentation on dark/light modes at some point. I reviewed but wasn't able to try it out because the branch is a ways behind `main`, but I can do that when it is merged in.

apps/web/src/styles/globals.css
dallashuggins: Ohhh right, that makes sense 💯

afraser: re: documenting the colors... yeah, my biggest gripe with this implementation is that these variables aren't even a proper css color since they need to be interpolated. This makes telling what they are using your editor (or viewing/modifying them in chrome devtools) is pretty meh.

Here's the tailwind docs on it FWIW:
https://tailwindcss.com/docs/customizing-colors#using-css-variables

dallashuggins: Totally, that was my thought as well. I did see that was Tailwind's recommendation though, and it is a clean approach, just would be nice if you could list the actual colors.

afraser: 👍🏼 working on getting main merged in now and then I'll update the ones that aren't just some shade of gray

apps/web/src/contexts/theme-context.tsx
```diff
@@ -0,0 +1,91 @@
...
+export const ThemeProvider = ({ children }: { children: React.ReactNode }) => {
+  const [theme, setTheme] = useState<string | undefined>(DEFAULT_THEME)
+
+  const toggleTheme = () => {
+    setTheme((theme) => (theme === 'dark' ? 'light' : 'dark'))
+  }
+
+  const handleThemeChangeShortcut = useCallback((event: KeyboardEvent) => {
+    if (event.ctrlKey && event.key === 'l') {
+      console.log('toggle theme')
```
smonn, line 55: I assume this is meant to be removed?

```suggestion
```

smonn (approved):
So much better with semantic names for all of those colors. Thanks for this! Just the one missed `console.log`, other than that, looks good to me.

PR #215, opened 2022-01-20
Auction smart contract in PyTEAL

### LINKED ISSUE #209: Implement marketplace smart contract

smonn:
Closes #209 

Initial implementation of the auction smart contract. Includes built files in TEAL for convenience when used in the backend. With some slight tweaks, this is essentially the same as the [auction demo](https://github.com/algorand/auction-demo) contract. Reasoning being that we don't need a much more custom setup than that once we remove the default frozen option from our asset creation transaction.


jakerainis (approved):
Amazing work. I'd love to talk through a couple of logistics with you for strategy purposes, but this is a great start! :ship-it:

smonn:
Also worth pointing out that this does not include any form of royalty mechanism. We'll also need to add a separate one if we want to support USDC as the currency (instead of ALGO).

davidjmurphyjr (approved):
I'm not proficient with any of this tech, but looks pretty good to me 👍

dallashuggins (approved):
Awesome! 🔥 I need to get more familiar with Python and smart contracts, but everything makes to my understanding.

PR #221, opened 2022-01-21
Upgrade three and fix dark styles

smonn:
Upgrade three to 0.136.0 and adjust some dark mode styles (transfer modal, pack opening, and untransferred banner)


jakerainis (approved):

jakerainis:
It's always good when Three upgrades work.

PR #222, opened 2022-01-21
Batch upgrade minor and patch dependencies

smonn:
As always, please try this locally and verify that the package-lock.json file does not change on your end. Let me know!


apps/web/src/utils/logger.ts
```diff
@@ -29,7 +29,7 @@ export const logger = pino({
 })
 
 export const loggerHttp = http({
-  logger,
+  logger: logger as Logger,
```
smonn, line 32: Some weird type incompatibility going on here. Still logs just fine.

jakerainis (approved):

dallashuggins (approved):
Everything working the same for me 💯

dallashuggins:
Oops, almost missed that the package-lock.json was regenerated for me.

PR #223, opened 2022-01-21
Admin interface proof of concept

### LINKED ISSUE #206: Proof of concept; Firebase claims + role-based auth page

dallashuggins:
### Changes
- New `FIREBASE_ADMIN_EMAIL` env variable. If set, when an account is created that matches this email, the admin custom claim will be automatically added to their Firebase user, allowing them to access admin-specific pages. 
  - It seemed preferential to do this on account creation, that way the passcode and username don't also have to be set as environmental variables, and the account can be created normally using Google or email and then permissions are added. Let me know if you have any thoughts there!
- Fix for "are you sure you want to leave?" prompt on form validation errors; added prompt state to payment context so it can be used for other payment methods eventually as well. 
- New `AdminService` with method for getting the logged-in user's permissions/claims. This is a separate service from auth because we are forcing a refresh when retrieving the token (i.e., `getIdToken(true)`). 
- New `isAuthenticatedUserAdmin` method in the api auth service, for use in the server side props on protected pages. 
  - We have a server side function to check the permissions (i.e., `isAuthenticatedUserAdmin`) _and_ a verification on page render (i.e., `getLoggedInUserPermissions`). This is to prevent Firebase caching of the claims. 
- Admin middleware that checks permissions, for use with any admin endpoints. 
- New endpoints for getting and setting claims which use admin middleware. The admin middleware checks if the user has permissions OR for the admin email. 
- Scaffolding for Transactions page, which only allows a logged-in user with permissions to access. 
- New types and validation

### Testing
- Set `FIREBASE_ADMIN_EMAIL` locally to the email you'll use to login
- Start application
- Create new account for this email (if Google, can delete user locally and sign in again)
- Try to access [the transactions page](http://localhost:3000/admin/transactions)

Closes #206


apps/web/src/pages/api/v1/admin/get-claims.ts
```diff
@@ -0,0 +1,26 @@
...
+handler.get(async (request: NextApiRequestApp, response: NextApiResponse) => {
+  if (!request.user.externalId) {
+    throw new BadRequest('No external ID provided')
+  }
+  const admin = configureAdmin()
+  const firebaseUser = await admin.auth().getUser(request.user.externalId)
+  const claims = firebaseUser.customClaims
+    ? Object.keys(firebaseUser.customClaims)
+    : []
+  response.status(200).json({ roles: claims })
```
afraser, line 23: It's a little confusing that `get-claims` returns `roles`. I think we should either return the claims as they are returned to us by firebase or rename this to `get-roles`.

apps/web/src/pages/api/v1/admin/update-claims.ts
```diff
@@ -40,12 +27,25 @@ handler.patch(
...
-      .auth()
-      .setCustomUserClaims(request.user.externalId, {
-        [request.body.role]: true,
-      })
+    const firebaseUser = await admin.auth().getUser(request.user.externalId)
+    const claims = firebaseUser.customClaims
+
+    // If the user is not admin OR the logged in user isn't the admin email, throw error
+    const isAdminUser =
+      !!claims && Object.keys(claims).includes(FirebaseClaim.admin)
```
afraser, line 31: How about `!!claims && (FirebaseClaim.admin in claims)`? or `claims?.[FirebaseClaim.admin]`

apps/web/src/pages/index.tsx
```diff
@@ -14,13 +16,29 @@ interface HomeProps {
...
+        console.log('claims:', claims)
+      } catch (error) {
+        console.error(error)
+      }
+    }
+    // Check permissions on page render, after auth token is refreshed so claims are fresh
+    if (auth.user) {
+      findUser()
+    }
+  }, [auth?.user])
```
afraser: If user is an object you may want to have this depend on something like `auth?.user.email`

afraser (approved):

apps/web/src/pages/api/v1/admin/update-claims.ts
dallashuggins: Good idea! Much simpler. Updated in 6b0a58d

apps/web/src/pages/admin/transactions.tsx
```diff
@@ -20,7 +20,7 @@ export default function AdminTransactionsPage() {
       try {
         const claims = await adminService.getLoggedInUserPermissions()
         // If there is no admin role, throw error
-        if (!claims.roles || !claims.roles.includes(FirebaseClaim.admin)) {
+        if (!claims.claims || !claims.claims.includes(FirebaseClaim.admin)) {
```
afraser, line 23: This is a little cleaner and doesn't involve iteration
```ts
const { claims } = await adminService.getLoggedInUserPermissions()
if (!claims?.[FirebaseClaim.admin]) { ... }
```

dallashuggins: Good call, just added in 59f3195.

afraser: Ah, just realized that claims here don't have the same structure as we have on the backend. I was thinking claims.claims was an object. Anyway, this all seems good to merge at this point to me. We should talk about what's next.

apps/web/src/pages/api/v1/profile/index.ts
```diff
@@ -42,6 +44,26 @@ handler.post(
       throw new BadRequest('Could not create account')
     }
 
+    // Check if the account is the admin account, and if so set admin permissions
+    if (body.email === Environment.firebaseAdminEmail) {
```
smonn, line 48: Does this mean only a single user can be admin?

apps/web/src/pages/api/v1/profile/index.ts
```diff
@@ -42,6 +44,26 @@ handler.post(
...
+          !firebaseUser.customClaims ||
+          !firebaseUser.customClaims[FirebaseClaim.admin]
+        ) {
+          await admin.auth().setCustomUserClaims(request.token.uid, {
+            [FirebaseClaim.admin]: true,
+          })
+        }
+      } catch (error) {
+        // do nothing if setting claims does not work
+        console.error(error)
```
smonn, line 63: Import `logger` from `@/utils/logger` and then change this line to:

```suggestion
        logger.error(error)
```

apps/web/src/middleware/admin-middleware.ts
```diff
@@ -0,0 +1,41 @@
...
+
+      // Check the verified claims of the token include admin permissions
+      const admin = configureAdmin()
+      const firebaseUser = await admin.auth().getUser(request.user.externalId)
+      const claims = firebaseUser.customClaims
+
+      // If the user is not admin, throw an error
+      const isAdminUser = claims?.[FirebaseClaim.admin]
+      if (!isAdminUser) {
+        throw new Unauthorized('User does not have admin permissions')
```
smonn, line 32: I _think_ we should be throwing a [`403 Forbidden`](https://httpstatuses.com/403) error here.

smonn (commented):
Some minor things and a question I'd like to clear up first!

apps/web/src/pages/api/v1/profile/index.ts
dallashuggins: I assumed we would be supporting adding other admin users in the near future, and there is an endpoint for updating the claims ready to be used, but I wasn't sure how we should display that to the admin user. I'm going to connect with Tom on that.

apps/web/src/pages/api/v1/profile/index.ts
smonn: May also wanna add some additional text to be logged, for example:

```suggestion
        logger.error(error, 'Unable to set claims')
```

apps/web/src/middleware/admin-middleware.ts
dallashuggins: Agreed, updated with 2de662d

apps/web/src/pages/api/v1/profile/index.ts
dallashuggins: Aw yes, thanks for catching! I updated with 2de662d.

smonn (approved):

jakerainis (commented):
Sorry for being so late to the party. Great work! 🙌

Issue #216, opened 2022-01-21
Update backend and data model to interact with smart contract for P2P auctions


smonn:
Proposed model, based on @davidjmurphyjr's comment in #212:

## CollectibleAuction

- `id` _guid_ - primary key
- `collectibleId` _guid_ - foreign key to a collectible
- `userAccountId` _guid_ - foreign key to a user account (represents the seller)
- `reservePrice` _int_ - minimum bid amount in microAlgos
- `startAt` _datetime with tz_ - start date + time bids are allowed to be made
- `endAt` _datetime with tz_ - end date + time bids are no longer allowed and auction is closed
- `createdAt` _datetime with tz_ - date listing was created
- `status` _active | closing | closed | canceled_ - some statuses to help with background tasks
- `appId` _int_ - the blockchain app (smart contract) ID
- `activeBidId` _guid_ - optional field for the current top bid
- `transactionId` _guid_ - foreign key to an algorand transaction for the app creation transaction

## CollectibleAuctionBid

- `id` _guid_ - primary key
- `collectibleAuctionId` _guid_ - foreign key to a listing
- `amount` _int_ - bid in microAlgos
- `userAccountId` _guid_ - foreign key to a user account (represents the bidder)
- `createdAt` _datetime with tz_ - date bid was made
- `transactionId` _guid_ - optional foreign key to an algorand transaction for the payment or app call transaction

smonn:
Was meant to be closed via #227

PR #227, opened 2022-01-21
Add marketplace schema and models

smonn:
Add corresponding schema and models for the migrations for #216

### DIFF
--- a/apps/api/src/models/collectible-auction.model.ts
+++ b/apps/api/src/models/collectible-auction.model.ts
@@ -1,0 +1,65 @@
+import {
+  CollectibleAuctionSchema,
+  CollectibleAuctionStatus,
+} from '@algomart/schemas'
+import { Model } from 'objection'
+
+import { AlgorandTransactionModel } from './algorand-transaction.model'
+import { BaseModel } from './base.model'
+import { CollectibleModel } from './collectible.model'
+import { CollectibleAuctionBidModel } from './collectible-auction-bid.model'
+import { UserAccountModel } from './user-account.model'
+
+export class CollectibleAuctionModel extends BaseModel {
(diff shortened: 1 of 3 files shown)


apps/api/src/models/collectible-auction.model.ts
```diff
@@ -0,0 +1,65 @@
...
+  reservePrice!: number
+  startAt!: string
+  status!: CollectibleAuctionStatus
+  transactionId!: string
+  userAccountId!: string
+
+  bids?: CollectibleAuctionBidModel[]
+  collectible?: CollectibleModel
+  transaction?: AlgorandTransactionModel
+  userAccount?: UserAccountModel
```
dallashuggins, line 29: Should the option to allow bid expiration be here, so that the next user would be awarded the collectible if it matched the reserve price, if the first bid payment failed? Wondering if the functionality should be automatic, or if it should be an option for the user, and/or if perhaps the user may want to accept/deny at the end of an auction instead of an automatic process. We could always add down the road though.

dallashuggins (approved):
Looks great!

libs/schemas/src/collectibles.ts
```diff
@@ -15,6 +15,13 @@ export enum IPFSStatus {
   Stored = 'stored',
 }
 
+export enum CollectibleAuctionStatus {
```
davidjmurphyjr, line 18: I think we might need an "Upcoming" status, should be added here, or would we compare now and startAt?

davidjmurphyjr (approved):
Looks good!

apps/api/src/models/collectible-auction.model.ts
dallashuggins: Oh right, awesome!

PR #239, opened 2022-01-25
Tables

afraser:
CleanShot 2022-01-25 at 13 00 58@2x
CleanShot 2022-01-25 at 13 01 10@2x


apps/web/src/components/credit-cards/credit-card.tsx
```diff
@@ -6,21 +6,22 @@ import css from './card.module.css'
...
   helpText: string
   href: string | { pathname: string; query?: { [key: string]: string } }
   icon: ReactNode
   isDisabled: boolean
   title: string
 }
 
-export default function Card({
+/** A credit card */
+export default function CreditCard({
```
afraser, line 18: I spent some time messing around with this component before realizing it wasn't a "card" in the sense that some UI libraries use.

apps/web/src/components/panel/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+export interface PanelProps {
+  className?: string
+  contentRight?: React.ReactNode
+  description?: React.ReactNode
+  footer?: React.ReactNode
+  fullWidth?: boolean
+  title?: React.ReactNode
+}
+
+export default function Panel({
```
afraser, line 16: We could also call this a `Card` if people prefer. ¯\_(ツ)_/¯

apps/web/src/components/credit-cards/credit-card.tsx
dallashuggins: Totally cool with renaming and/or reworking the components, but this component isn't specific to credit cards. This is used for the methods on the checkout page, which will include credit cards, wire payments, and crypto payments if all are enabled.

afraser: Ahhh crud. That makes sense though! Ok I'll have to take a closer look at this. I don't think it makes sense to have a global card components that doesn't accept `children`. Do you think it would make sense to move it into `components/purchase-form`? Or actually, looking at this now, maybe we just rename it to something like `PaymentOptionItem` and `PaymentOptions`. Thoughts?

afraser:
@dallashuggins I pulled your closed issues into this PR description but from our convo this morning it sounds like this doesn't in fact close #241. Can you confirm?

dallashuggins:
> @dallashuggins I pulled your closed issues into this PR description but from our convo this morning it sounds like this doesn't in fact close #241. Can you confirm?

@afraser Sweet, thanks! Re: closing #241 - good call, I'm trying to think what makes sense. I had been thinking that ticket was just back-end functionality, but if we were going to implement the search in the UI later, than this could be left as not complete until that's finished. Is that what you were thinking? Also, we could add searching by pack title in addition, but I could also see that being a requested feature down the road if its found useful.

afraser:
@dallashuggins Oh, this is just a case of "I'm not using it therefore it doesn't exist." I see the functionality now and agree that the issue should just be for the backend.

apps/web/src/hooks/use-pagination.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  const setPage = (page) => setPagination({ page })
+  const setSortBy = (sortBy) => setPagination({ sortBy })
+  const setSortDirection = (sortDirection) => setPagination({ sortDirection })
+
+  const handleSortBy = (newSortBy) => {
+    const nextSortDirection =
+      pagination.sortBy === newSortBy
+        ? SORTS[(SORTS.indexOf(pagination.sortDirection) + 1) % 3]
```
dallashuggins, line 27: Trying to understanding what's happening here - is dividing by 3 required due to the sort options? Should this be divided by `SORTS` length?

dallashuggins: I was able to sort successfully, so I know it works, just curious :)

dallashuggins (approved):
Nice work! 🔥 The table looks/works great - was able to try out pagination and sorting of the payments I have locally. Just added a couple clarifying comments.

apps/web/src/components/table/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+import css from './table.module.css'
+
+export type ColumnDefinitionType<T> = {
+  key: string
+  name: string
+  sortable?: boolean
+  // tooltip?: React.ReactNode
+  // stickLeft?: boolean
+  // stickRight?: boolean
+  // width?: number
```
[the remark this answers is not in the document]
afraser: yeah, leaving these in for future work

PR #240, opened 2022-01-26
[#214] Payments list

### LINKED ISSUE #214: API endpoint for listing payments (filtering, sorting, pagination)

### LINKED ISSUE #241: Search functionality for payments list endpoints

dallashuggins:
### Changes
- Payments list endpoint for querying for payments
  - Pagination through payments
  - Filter payments by pack ID, pack slug, or the payer's username or external ID
    - Started building filtering by pack title, but whenever I'd apply the translation filter, it would result in the translations not being returned in the response body, and it wouldn't filter the results. Since we can filter by the pack slug, holding on this functionality in the interest of moving faster, but let me know if you think this is very important to have. 
  - Sort by created at or updated at
  - Plumbing in web for requesting payments
- Making initial protected request to get payments on [Transactions page](http://localhost:3000/admin/transactions)
  - You can console log payments and total on the Transactions page, returned from server side props. We will need to filter, sort, and paginate when the user interacts with the page, so we'll need to utilize the checkout service method when adding to the page. 

Closes #214 
Closes #241


apps/api/src/modules/payments/payments.routes.ts
```diff
@@ -23,7 +24,12 @@ export async function getPublicKey(
   const paymentService = request
     .getContainer()
     .get<PaymentsService>(PaymentsService.name)
-  reply.send(await paymentService.getPublicKey())
+  const publicKey = await paymentService.getPublicKey()
+  if (publicKey) {
+    reply.send(publicKey)
+  } else {
+    reply.notFound()
+  }
```
dallashuggins, line 32: This fixes the issue where the public key endpoint was failing but it showed as successful

PR #246, opened 2022-01-27
Minor tweaks to card payments | Unclaimed logic

IanPirro:
These are a few things I noticed during a client launch.

### DIFF
--- a/apps/api/src/modules/packs/packs.service.ts
+++ b/apps/api/src/modules/packs/packs.service.ts
@@ -25,6 +25,7 @@ import {
   PackType,
   PackWithCollectibles,
   PackWithId,
+  PaymentStatus,
   PublishedPack,
   PublishedPacksQuery,
   SortDirection,
@@ -41,6 +42,7 @@ import { BidModel } from '@/models/bid.model'
 import { CollectibleModel } from '@/models/collectible.model'
 import { EventModel } from '@/models/event.model'
 import { PackModel } from '@/models/pack.model'
+import { PaymentModel } from '@/models/payment.model'
 import { UserAccountModel } from '@/models/user-account.model'
 import AccountsService from '@/modules/accounts/accounts.service'
 import CollectiblesService from '@/modules/collectibles/collectibles.service'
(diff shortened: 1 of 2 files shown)


apps/api/src/modules/packs/packs.service.ts
```diff
@@ -749,7 +751,22 @@ export default class PacksService {
     const allPacks: PackByOwner[] = []
     for (const { id, claimedAt, templateId } of packs) {
       const template = templateLookup.get(templateId)
-      if (template && claimedAt) {
+      // For packs that require payment, we need to check if the user has paid
```
IanPirro, line 754: This is to fix for a case where people where able to claim (transfer to their custodial) before their payment was marked as confirmed or paid.

apps/api/src/modules/payments/payments.service.ts
```diff
@@ -745,7 +745,7 @@ export default class PaymentsService {
       pendingPayments.map(async (payment) => {
         let status: PaymentStatus | undefined
         // Card flow
-        if (payment.externalId && payment.paymentCardId) {
+        if (payment.externalId) {
```
IanPirro, line 748: Checking for payment.paymentCardId here causes cards a user did not select to save for later use to never have their payment status updated. (In production every payment comes in as pending, even if it can be updated a few seconds later).

apps/api/src/modules/payments/payments.service.ts
```diff
@@ -759,6 +759,20 @@ export default class PaymentsService {
               status: circlePayment.status,
             })
             status = circlePayment.status
+            // If the new status is a final failed status, release the pack
```
IanPirro, line 762: If we are changing a payment from pending to failed, release the pack. The logic where we make create the payment in circle doesn't catch this because we check if the payment could not be created, but even if it fails the payment is created as pending and later updated to failed in this task.

apps/api/src/modules/packs/packs.service.ts
IanPirro: This logic actually needs to apply to auctions too I think, though I didn't look to see if it's the same

smonn: I think you're right, and my guess is it needs to be exactly the same for auctions. However, I'm not sure if any of this is different for crypto and wire payments. @dallashuggins?

smonn: I also made these changes in my PR to avoid some edge cases: https://github.com/deptagency/algomart/pull/155/files#diff-9163417dd8355447ea09b0ec98fb195305136b1506876ef2c02ddfbecde5a1ceL720-R733

IanPirro: great, we can just address these specific changes in your PR

PR #247, opened 2022-01-27
Update documentation

davidjmurphyjr:
A few small updates to the docs that I thought might help with future onboarding.

### DIFF
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -62,7 +62,7 @@ CURRENCY=
 COINBASE_URL=https://api.coinbase.com
 
 # pinata.cloud configuration (for storing on IPFS)
-# sign up for a free tier pinata.cloud account using your Rocket email
+# You can sign up for a free tier https://pinata.cloud account with 1 GB storage to start
 PINATA_API_KEY=
 PINATA_API_SECRET=
 
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -13,7 +13,10 @@ ALGOD_PORT=
 ALGOD_TOKEN=
 ALGOD_ENV=testnet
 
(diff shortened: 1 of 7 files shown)


apps/api/.env.example
```diff
@@ -13,7 +13,10 @@ ALGOD_PORT=
 ALGOD_TOKEN=
 ALGOD_ENV=testnet
 
-# CMS configuration (CMS_ACCESS_TOKEN should match the KEY environment variable in the CMS .env)
+### CMS configuration
+# CMS_ACCESS_TOKEN should match...
+# - the KEY environment variable in the CMS .env
+# - the Admin User > Admin Options > Token field in the CMS UI
```
smonn, line 19: Sorry, this is incorrect (the old one was wrong too). This should **not** match the `KEY` environment variable from the CMS. That is just used for the CMS session cookie I believe. `CMS_ACCESS_TOKEN` should **only** match the latter (Admin User > Admin Options > Token field in the CMS UI).

apps/api/.env.example
```diff
@@ -59,6 +62,7 @@ CURRENCY=
 COINBASE_URL=https://api.coinbase.com
 
 # pinata.cloud configuration (for storing on IPFS)
+# sign up for a free tier pinata.cloud account using your Rocket email
```
smonn, line 65: Not Rocket 😉 

```suggestion
# You can sign up for a free tier https://pinata.cloud account with 1 GB storage to start
```

smonn (commented):
Thanks! Some small tweaks are needed, but looks good.

apps/api/.env.example
davidjmurphyjr: This process was documented in the some of the main and CMS READMEs. I moved it to the API README, as that seemed like the right place for it.

smonn (approved):
Thanks for all of this, LGTM!

PR #251, opened 2022-01-27
Upgrade tailwind to lts (^3.0.17)

larpollack:
Upgraded tailwind and its dependencies & configs; fixed any deprecation bugs.

Manually tested all pages in light and dark mode.


package.json
```diff
@@ -138,16 +138,17 @@
...
     "react-test-renderer": "17.0.2",
     "sqlite3": "5.0.2",
-    "tailwindcss": "2.2.19",
+    "tailwindcss": "^3.0.17",
     "ts-jest": "27.1.3",
     "ts-node": "10.4.0",
     "tsconfig-paths": "3.12.0",
-    "typescript": "4.5.5"
+    "typescript": "4.5.5",
+    "undefined": "^0.1.0"
```
smonn, line 151: Is this`"undefined"` module meant to be here?

package.json
```diff
@@ -98,8 +98,8 @@
     "@nrwl/tao": "13.4.6",
     "@nrwl/web": "13.4.6",
     "@nrwl/workspace": "13.4.6",
-    "@tailwindcss/aspect-ratio": "0.4.0",
-    "@tailwindcss/line-clamp": "0.3.0",
+    "@tailwindcss/aspect-ratio": "^0.4.0",
```
smonn, line 101: Typically we stick with fixed versions in the `package.json` file. So please remove the `^` here and below.

```suggestion
    "@tailwindcss/aspect-ratio": "0.4.0",
```

package.json
larpollack: I have no idea what that is or how it got there. Will remove

package.json
```diff
@@ -99,7 +99,7 @@
     "@nrwl/web": "13.4.6",
     "@nrwl/workspace": "13.4.6",
     "@tailwindcss/aspect-ratio": "0.4.0",
-    "@tailwindcss/line-clamp": "0.3.0",
+    "@tailwindcss/line-clamp": "^0.3.1",
```
afraser, line 102: Everything looks good except it looks like we need to pin these. I think you'll need to run `npm i` for each of these without the `^` (so `npm i @tailwindcss/line-clamp@0.3.1` in this case) in order to make sure package.lock is updated accordingly.

afraser (approved):
pin those package versions and :shipit:

package.json
smonn: I think these are dev deps, so `npm i --save-dev package-name@1.2.3` will be needed.

smonn (approved):
:shipit:  🤞

PR #264, opened 2022-01-31
Readme typos, clarity. Better directus executor failing

tj-boyle:
Fix of various typos and missing clarity of READMEs.

Added an additional step to CMS setup regarding creating a DB schema that matches the `DB_SEARCH_PATH` environment variable. The env example has `cms` as a value there, so if you keep that value without creating a schema, `nx bootstrap cms` will fail.

BUT it will tell you it succeeded - so I additionally modified the executor to throw an error if there's an error instead of always returning success.

May also modify bottom portion of the CMS Readme before merge, see: https://rocketinsights.slack.com/archives/C01RTKC55R6/p1643668237913089

### DIFF
--- a/tools/executors/directus/impl.js
+++ b/tools/executors/directus/impl.js
@@ -193,7 +193,7 @@ function directusExecutor(options, context) {
           return [4 /*yield*/, runAction(options, context)]
         case 1:
           _a.sent()
-          return [2 /*return*/]
+          return [2 /*return*/, { success: true }]
       }
     })
   })
--- a/tools/executors/directus/impl.js
+++ b/tools/executors/directus/impl.js
@@ -177,6 +177,7 @@ function runAction(options, context) {
           directus.on('exit', function (code) {
             if (code !== 0) {
               reject(code)
+              return
             }
             resolve(true)
           })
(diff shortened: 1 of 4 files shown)


tools/executors/directus/impl.js
```diff
@@ -174,7 +174,10 @@ function runAction(options, context) {
           directus.on('error', function (error) {
             reject(error)
           })
-          directus.on('exit', function () {
+          directus.on('exit', function (code) {
+            if (code !== 0) {
+              reject(code)
```
smonn, line 179: I _think_ we need a `return` statement here too.

```suggestion
              reject(code)
              return
```

tools/executors/directus/impl.ts
```diff
@@ -45,7 +45,11 @@ async function runAction(
       reject(error)
     })
 
-    directus.on('exit', () => {
+    directus.on('exit', (code) => {
+      if (code !== 0) {
+        reject(code)
```
smonn, line 50: Same as above.

```suggestion
        reject(code)
        return
```

smonn (approved):
Thanks for this!

tools/executors/directus/impl.js
tj-boyle: yep turns out. the promise stays rejected but it'd still execute any other code afterward if not returned, never really even considered that!

smonn: Now that I think about it, another option is to wrap the resolve in an else block.

smonn (approved):

PR #268, opened 2022-02-01
[#255, 267, 270] Admin transactions endpoints

### LINKED ISSUE #255: API endpoint for searching for payments matching bank account ID in Circle

### LINKED ISSUE #267: API endpoint for getting an individual payment with admin details

### LINKED ISSUE #270: API endpoint for updating a payment

dallashuggins:
### Changes
- Built on top of #239 
- Endpoint for searching for all wire payments by the bank account ID, for use on the individual Transaction page, when retrieving all payments to display in the table.
- Endpoint for getting an individual payment, for use on the individual Transaction page
- Endpoint for updating a payment
  - Example use on the Transaction page, showing how we would reset to pending and update to paid. There's a placeholder button for revoking even if that's not done yet. 
  - If a payment is made with a wrong amount, nothing would be done in the interim except confirm the state (you wouldn't want to reset the associated payment since its valid), and the customer service rep could return back, and if the other payment was shown they could mark as Paid. 

Closes #255 
Closes #267 
Closes #270


apps/api/src/lib/circle-adapter.ts
```diff
@@ -352,14 +352,6 @@ export default class CircleAdapter {
...
   ): Promise<ToPaymentBase | null> {
     const searchParams = {}
-    if (query.walletId)
-      Object.assign(searchParams, { walletId: query.walletId })
-    if (query.sourceWalletId)
-      Object.assign(searchParams, { sourceWalletId: query.sourceWalletId })
-    if (query.destinationWalletId)
-      Object.assign(searchParams, {
-        destinationWalletId: query.destinationWalletId,
-      })
```
dallashuggins, line 362: No longer being used, since we're querying the Core API, not the Accounts API.

apps/web/src/pages/admin/transactions/[transactionId].tsx
```diff
@@ -0,0 +1,114 @@
...
+    const paymentId = typeof transactionId === 'string' ? transactionId : null
+    const updatedPayment = await adminService.updatePayment(paymentId, {
+      status: PaymentStatus.Paid,
+    })
+    console.log('marked payment as paid:', updatedPayment)
+  }, [transactionId])
+
+  const handleRevokePack = useCallback(async () => {
+    // @TODO: Revoke pack API request
+  }, [])
```
Comment (dallashuggins) [inline apps/web/src/pages/admin/transactions/[transactionId].tsx:51]: Examples requests that can be made on this page.

apps/web/src/services/checkout-service.ts
```diff
@@ -141,6 +141,16 @@ export class CheckoutService implements CheckoutAPI {
...
+  async getPaymentsByBankAccountId(
+    bankAccountId: string
+  ): Promise<AdminPaymentList> {
+    return await this.http
+      .get(
+        `${urls.api.v1.admin.getPaymentsForBankAccount}?bankAccountId=${bankAccountId}`
+      )
+      .json<AdminPaymentList>()
+  }
+
```
dallashuggins, line 153: Should this be in the admin service? I was conflicted where to place these functions. The admin middleware is used for these endpoints, so maybe it'd be worth keeping in admin.

apps/api/src/modules/payments/index.ts
```diff
@@ -97,6 +102,20 @@ export async function paymentRoutes(app: FastifyInstance) {
       },
       getPaymentById
     )
+    .get(
+      '/:paymentId/admin',
```
afraser, line 106: Is there a reason not to nest all admin-only api routes under a top level admin/ namespace? Tacking admin on the end doesn't feel in keeping with rest conventions.

apps/web/src/services/checkout-service.ts
afraser: I would think so, yes.

afraser (commented):
This is looking good although we're starting to step on each others toes. I'm going to go ahead and merge this branch into my own so I can resolve some of the conflicts and test stuff out. Looking good though!

apps/api/src/modules/payments/index.ts
dallashuggins: As far as the back-end is concerned they aren't admin routes, though we could still organize that way. I was unsure how to deal with this route, as it has additional information returned, but otherwise is just a normal endpoint for getting payment details. I added "admin" but that doesn't feel appropriate necessarily. Any thoughts on how we should organize?

afraser: Ah I gotcha. Good call. Hrm. This is probably fine then. The only other thing I can think of is something like "with-x" if all it did was return some additional relation.

afraser (approved):

apps/api/src/modules/payments/index.ts
dallashuggins: Aw yeah, good point. That makes more sense - I can convert to that structure.

dallashuggins:
Closing since #271 will contain the work done in this PR

PR #271, opened 2022-02-01
admin transaction page

afraser:
CleanShot 2022-02-04 at 16 20 17@2x

----

CleanShot 2022-02-04 at 16 19 52@2x

----

CleanShot 2022-02-04 at 09 27 13@2x


apps/web/src/components/step-links/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export interface StepLinksProps {
+  links: {
+    label: string
+    href: string
+    handleClick?: () => void
+  }[]
+}
+
+/** */
```
dallashuggins, line 19: Were you going to leave a comment here?

apps/web/src/layouts/admin-layout.tsx
```diff
@@ -0,0 +1,36 @@
...
+import EmailVerification from '@/components/profile/email-verification'
+import UntransferredPacks from '@/components/profile/untransferred-packs'
+
+export interface DefaultLayoutProps {
+  children?: ReactNode
+  pageDescription?: string
+  pageTitle?: string
+}
+
+export default function DefaultLayout({
```
dallashuggins, line 17: Should this be `AdminLayout`?

apps/web/src/hooks/use-admin-gate.ts
```diff
@@ -0,0 +1,35 @@
...
+        console.error(error)
+        router.push(urls.home)
+      }
+    }
+
+    // Check permissions on page render, after auth token is refreshed so claims are fresh
+    if (auth.user) {
+      verifyAdminUser()
+    }
+  }, [auth?.user, router])
```
dallashuggins, line 32: 🔥

dallashuggins (approved):
The individual transaction page and the transactions list page look amazing! 🙌 Everything is working as expected for me. 🔥

apps/web/src/pages/admin/transactions/[transactionId].tsx
```diff
@@ -0,0 +1,288 @@
...
+      logger.info(updatedPayment, 'Payment marked as paid.')
+    } catch (error) {
+      alert('Unable to update pack as paid.')
+      logger.error(error, 'Unable to update pack as paid')
+    }
+  }, [transactionId])
+
+  const handleRevokePack = useCallback(async () => {
+    if (!confirm('Are you sure you want to revoke this pack?')) return
+    // @TODO: Revoke pack API request
```
Comment (IanPirro) [inline apps/web/src/pages/admin/transactions/[transactionId].tsx:96]: Is this TODO still necessary?

apps/api/src/modules/packs/packs.service.ts
```diff
@@ -854,6 +855,77 @@ export default class PacksService {
...
+            userId,
+            trx
+          ))
+      )
+    )
+
+    // Create transfer success notification to be sent to user
+    const packWithBase = await this.getPackById(request.packId)
+    if (packWithBase) {
+      // @TODO: create notification for revoking the pack
```
IanPirro, line 907: Is this for a later pr?

IanPirro (approved):

apps/web/src/pages/admin/transactions/[transactionId].tsx
Comment (afraser) [reply apps/web/src/pages/admin/transactions/[transactionId].tsx:96]: nope!

apps/api/src/modules/packs/packs.service.ts
afraser: yeah

apps/api/src/modules/packs/index.ts
```diff
@@ -208,6 +210,23 @@ export async function packsRoutes(app: FastifyInstance) {
     mintPackStatus
   )
 
+  app.post(
```
davidjmurphyjr, line 213: This is an admin-only endpoint and it's living alongside end-user endpoints, I don't see a check to make sure the  the caller is an admin (but I def could be overlooking it). Do we trust the web layer to do the [appropriate authorization](https://github.com/deptagency/algomart/pull/271/files#diff-fbddb1ab07ada423622f6f4c479e2ca80410238045d7d4209c2083835cde88e9R15) or does the API layer need it's own authorization?

davidjmurphyjr (approved):
Looks good to me! Left a comment, but it's more to educate myself about Algomart's conventions

apps/api/src/modules/packs/index.ts
dallashuggins: Currently all of the admin-specific endpoints are living among the other endpoints, and we are trusting the web layer to do the appropriate authorization. There is no concept of an admin role in the API, since Firebase Admin isn't used in the API and that's what we're using for determining permissions. It's definitely something to continue considering, if we want to add authorization to the API layer, but it didn't seem necessary at this time. Typically you would implement Firebase Admin in the API, but since the front-end has it's own API server, we can do the authorization on those endpoints using the new adminMiddleware instead. This seemed secure, as this verification needs to happen in order to call the endpoints, but ideally the API would determine this.

afraser: @dallashuggins this one's for you.

davidjmurphyjr: makes sense, thanks for explaining!

PR #272, opened 2022-02-01
[#269] Revoke pack endpoint

### LINKED ISSUE #269: API endpoint for revoking a pack claim

dallashuggins:
### Changes
- Collectible service method for transferring from the user account to the creator account.
- Clawback method for revoking from the user to the creator account
- Revoke pack method in the pack service that revokes the claim and transfers the asset back to the creator account. 
- [ ] Determining when to enable/disable buttons/functionality based on the current state of the payment. 
- [ ] User notification emails

Closes #269

### DIFF
--- a/apps/api/src/modules/payments/payments.service.ts
+++ b/apps/api/src/modules/payments/payments.service.ts
@@ -1,6 +1,4 @@
 import {
-  AdminPaymentList,
-  AdminPaymentListQuerystring,
   CirclePaymentQueryType,
   CirclePaymentSourceType,
   CreateBankAccount,
@@ -17,7 +15,9 @@ import {
   PackType,
   PaymentBankAccountStatus,
   PaymentCardStatus,
+  Payments,
   PaymentSortField,
+  PaymentsQuerystring,
   PaymentStatus,
   SendBankAccountInstructions,
   SortDirection,
@@ -79,7 +79,7 @@ export default class PaymentsService {
     payerUsername,
     sortBy = PaymentSortField.UpdatedAt,
     sortDirection = SortDirection.Ascending,
-  }: AdminPaymentListQuerystring): Promise<AdminPaymentList> {
+  }: PaymentsQuerystring): Promise<Payments> {
     let account: UserAccount
     userInvariant(page > 0, 'page must be greater than 0')
     userInvariant(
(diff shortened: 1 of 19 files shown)


apps/api/src/modules/payments/payments.service.ts
```diff
@@ -762,29 +762,23 @@ export default class PaymentsService {
...
+          templateIds: [pack.templateId],
+        })
+        const packTemplate = packTemplates[0]
+        return {
+          ...payment,
+          pack: {
+            ...packTemplate,
+            ...pack,
+          },
+        }
```
dallashuggins, line 781: @afraser I also pushed 68205f3 with another update we discussed, passing in a query parameter to determine what gets returned by the `getPaymentById` method, removing the extra method, and cleaning up mentions of "admin" in endpoint names.

dallashuggins: Still working on this revoke endpoint though, and planned to merge this PR into my other PR when complete, but let me know if I'm holding you up in any way and I can get these changes into the other branch.

afraser: @dallashuggins no hold up here. Your last PR came at the perfect time so I'm currently working on filling in the transaction show page with real data and styles.

apps/web/src/clients/api-client.ts
```diff
@@ -316,23 +316,21 @@ export class ApiClient {
...
+      .get(`payments/${paymentId}?isAdmin=${true}`)
+      .json<Payment>()
   }
 
   async getPaymentsByBankAccountId(bankAccountId: string) {
     return await this.http
-      .get(`bank-accounts/${bankAccountId}/payments`)
-      .json<Payment[]>()
+      .get(`payments/bank-accounts/${bankAccountId}/payments`)
+      .json<ToPaymentBase[]>()
```
dallashuggins, line 333: Accidentally was using the `Payment` type, which includes additional database the database such as the `payerId`, but we just need to return a list of payments from Circle, some of which could not exist in the database.

apps/web/src/utils/asset-validation.ts
```diff
@@ -72,6 +72,12 @@ export const validateClaimAsset = (t: Translate) =>
     packTemplateId: packTemplateId(t),
   })
 
+export const validateRevokeAsset = (t: Translate) =>
+  object({
+    packId: packId(t),
+    ownerId: string(required(t('forms:errors.required') as string)),
+  })
```
dallashuggins, line 79: This currently requires an `ownerId`, but the revoke pack schema includes a `fromAddress` that could be used optionally if we wanted to transfer from an address which wasn't connected to a user in the database. Though this is the front-end validation, which does require the owner ID for now.

dallashuggins:
Closing since the work here was merged into #271

PR #274, opened 2022-02-03
CMS translation defaults

tj-boyle:
Refactored the Directus adapter:
- No longer filters requests down to only user's locale
- Created new model for translation types, which each type extends from
- Response translations now call function to get translation
- Said function checks directus response for translation in user's locale, otherwise defaulting to english

Note: This is done on an object by object basis. This does _not_ default to english on a field by field basis. 
If the French translation has _some_ missing fields, the API will no longer throw errors but the UI will show blank data. IMO this is a good tradeoff for now.


apps/api/src/lib/directus-adapter.ts
```diff
@@ -371,20 +394,24 @@ export function toCollectionWithSets(
...
+    getDirectusTranslation<DirectusCollectibleTemplateTranslation>(
+      template.translations as DirectusCollectibleTemplateTranslation[],
+      `collectible ${template.id} has no translations`,
+      locale
+    )
 
   const rarity = template.rarity as DirectusRarity
-  const [rarityTranslation] = rarity?.translations || []
-  invariant(
-    rarityTranslation === undefined || typeof rarityTranslation !== 'number',
```
davidjmurphyjr, line 385: Not 100% sure what's going on here, I think the invariant logic is changing  from `rarityTranslation === undefined` to `translation !== undefined` via the call to `getDirectusTranslation`

Seems like `rarityTranslation === undefined ` might have been incorrect though...

davidjmurphyjr (approved):
LGTM

apps/api/src/lib/directus-adapter.ts
tj-boyle: Yeah I believe it was incorrect, it didn't make much sense as is

tj-boyle:
@afraser All the translations are typed to either be an array of numbers an array of objects - as for why is still somewhat beyond me, as prior to my changes all the checks were ensuring it _wasn't_ a number. 

I think if the Directus CMS Collection translations are improperly configured they come back as index numbers instead of strings - so potentially that's why the check is there, to ensure the CMS has been configured properly

afraser:
@tj-boyle Gotcha. Thanks for the explanation!

PR #285, opened 2022-02-08
Type error bug fix

dallashuggins:
### Changes
- Type fix to prevent "allowBidExpiration is required" error when calling the getting payment by ID endpoint

### DIFF
--- a/apps/web/src/pages/admin/transactions/[transactionId].tsx
+++ b/apps/web/src/pages/admin/transactions/[transactionId].tsx
@@ -33,7 +33,7 @@ export default function AdminTransactionPage({
   const { t, lang } = useTranslation('admin')
   const { query } = useRouter()
   const { transactionId } = query
-  const isAuction = !!payment.pack?.template.auctionUntil
+  const isAuction = !!payment.pack?.template?.auctionUntil
 
   // WIRE PAYMENTS
   const { data } = useAuthApi<WirePayment[]>(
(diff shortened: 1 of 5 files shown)


libs/schemas/src/packs.ts
```diff
@@ -214,6 +214,34 @@ export const PackByTemplateIdSchema = Type.Object({
...
+          slug: Type.String(),
+          status: Type.Enum(PackStatus),
+          subtitle: Type.Optional(Type.String()),
+          templateId: IdSchema,
+          title: Type.String(),
+          type: Type.Enum(PackType),
+        })
+      )
+    ),
+  }),
```
dallashuggins, line 242: Passing in an optional object with template details allows us to only send these extra details when the admin argument is used.

dallashuggins: Updated to use the existing `PackBaseSchema` type for `template`.

apps/web/src/pages/admin/transactions/[transactionId].tsx
```diff
@@ -33,7 +33,7 @@ export default function AdminTransactionPage({
   const { t, lang } = useTranslation('admin')
   const { query } = useRouter()
   const { transactionId } = query
-  const isAuction = !!payment.pack?.auctionUntil
+  const isAuction = !!payment.pack?.template.auctionUntil
```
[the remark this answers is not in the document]
Comment (dallashuggins) [reply apps/web/src/pages/admin/transactions/[transactionId].tsx:36]: The pack has a `template` object on it with the additional details retrieved by the get published packs endpoint, so the variable is correctly used, unless you mean we should change where the data is.

afraser (approved):

PR #290, opened 2022-02-17
Update Pack and NFT detail views

afraser:
- surface pack “additional images”
- consistently display title, subtitle, and body text (when present) for both the packs and nft detail views
- surfacing the Set (if present) and unique_code in the nft detail view metadata
- fixes some styles and layout quirks

Pack detail view
---
CleanShot 2022-02-17 at 12 41 23@2x
CleanShot 2022-02-17 at 12 41 33@2x

NFT detail view
---
CleanShot 2022-02-17 at 12 42 46@2x

Home page
---
CleanShot 2022-02-17 at 14 50 44@2x
CleanShot 2022-02-17 at 14 51 15@2x

Auctions
---
CleanShot 2022-02-17 at 15 21 36@2x
CleanShot 2022-02-17 at 15 21 44@2x


apps/api/src/lib/directus-adapter.ts
```diff
@@ -493,6 +493,9 @@ export function toPackBase(
 
   return {
     allowBidExpiration: template.allow_bid_expiration,
+    additionalImages: isStringArray(template.additional_images)
+      ? template.additional_images.map((id) => getFileURL(id))
+      : template.additional_images.map((t) => getFileURL(t.directus_files_id)),
```
davidjmurphyjr, line 498: Small thing, what does `t` stand for? Maybe use `ai` for additional_images or `df` for directus file?

apps/api/src/modules/payments/payments.service.ts
```diff
@@ -736,7 +737,54 @@ export default class PaymentsService {
...
+              amount: sourcePayment.amount,
+            },
+          },
+          trx
+        )
+      }
+    }
+
+    // Send email notification to  Customer service
+    if (Configuration.customerServiceEmail) {
```
davidjmurphyjr, line 758: This is the same as the check on line  741, I think the branching could be simplified.

apps/api/.env.example
```diff
@@ -30,6 +30,8 @@ CIRCLE_URL=https://api-sandbox.circle.com
 EMAIL_FROM=
 EMAIL_NAME=AlgoMart
 
+CUSTOMER_SERVICE_EMAIL=
```
davidjmurphyjr, line 33: I think it might be worth adding a comment explaining what this variable does. If I understand correctly, some clients want to their customer service people to receive email notifications and some don't, so `CUSTOMER_SERVICE_EMAIL` is optional, when it's provided they get emails, when it's not, they don't.

dallashuggins (approved):
Nice improvements! No additional comments from what David already mentioned about the naming and possibly simplifying the logic a bit. Tried out the revoke pack and it works well; I can see us adding a modal instead of the basic prompt, though not important. It gave me an option to block the alerts even, which I hadn't seen before.

davidjmurphyjr (approved):
LGTM

jakerainis (approved):

IanPirro (approved):
🎉

PR #296, opened 2022-02-24
Issue with errors / transact plugin

IanPirro:
Noticed this when implementing a non-algomart project with a near identical fastify setup including this custom transact plugin.

For routes that ` app.setErrorHandler(appErrorHandler(app))` (essentially all routes) the onError hook in the transact plugin will not get called, which results in the transaction not getting rolled back (it gets committed). This can be solved other ways, this is just how I chose to solve it for now. We could just return an object that extends Error

See comments in code.

### DIFF
--- a/apps/api/src/plugins/transaction.plugin.ts
+++ b/apps/api/src/plugins/transaction.plugin.ts
@@ -59,7 +59,17 @@ export default fp(async function fastifyTransaction(fastify: FastifyInstance) {
   }
 
   const onSend: onSendAsyncHookHandler<unknown> = async (request) => {
+    // Error was handled by default error handler -> Transaction has already been rolled back
     if (request.transactionFailed) return
+
+    // Error was handled by custom error handler -> rollback
+    if (request.customError) {
+      fastify.log.error(request.customError, 'rollback transaction')
+      await request.transaction?.rollback()
+      return
+    }
+
+    // Success -> commit
     fastify.log.info('commit transaction')
     await request.transaction?.commit()
   }
--- a/apps/api/src/utils/errors.ts
+++ b/apps/api/src/utils/errors.ts
@@ -3,6 +3,15 @@ import { FastifyError } from 'fastify-error'
 import { HTTPError } from 'got'
 import { HttpError } from 'http-errors'
 
+declare module 'fastify' {
+  interface FastifyRequest {
+    /**
+     * Indicates that the error was handled by custom error handler.
+     */
+    customError?: Error
+  }
+}
+
 export class UserError extends Error {
   constructor(message: string, readonly statusCode: number) {
     super(message)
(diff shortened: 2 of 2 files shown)


davidjmurphyjr:
Just to confirm I understand the problem being solved here.... If I wanted to test this what would I do? I'm guessing I'd have a route hander (that uses `app.setErrorHandler(appErrorHandler(app))`) do some DB work, then throw an Error. Before this change, I should expect the DB work to get committed, after this change I should expect the DB work to be rolled back?

---

I'm not expert here so take this with a grain of salt... I took a look at `appErrorHandler` and if i'm understanding it correctly it's not doing a whole lot more that what the [default error handler](https://github.com/fastify/fastify/blob/main/lib/error-handler.js#L68) does, the only added bits of functionality seem to be `if (error instanceof UserError) reply.status(statusCode)` and removing `code: String  // the Fastify error code` from the reply. I proposed we remove the duplicated functionality from `appErrorHandler` and allow the default error handler to call [reply.send(error) ](https://www.fastify.io/docs/latest/Reference/Reply/#errors)for us. In doing so allow, we allow transaction.plugin's [onError handler ](https://www.fastify.io/docs/latest/Reference/Hooks/#onerror) to get called.


> onError... This hook will be executed only after the customErrorHandler has been executed, and only if the customErrorHandler sends an error back to the user (Note that the default customErrorHandler always sends the error back to the user).

afraser:
Thanks for doing this @IanPirro . I wonder if this is actually related to the problem I was showing you @davidjmurphyjr regarding email notifications being sent despite failed invariants.

All things being equal, I'd say what David is proposing would be better, but if the current implementation gives us any added assurances or flexibility then I'm all for it.

IanPirro:
@davidjmurphyjr 

Correct, in accounst.service you can change `userInvariant(!existing, 'username or externalId already exists', 400)` to `userInvariant(existing, 'username or externalId already exists', 400)` and then use swagger to submit any request to it.

Before
Screen Shot 2022-02-24 at 6 39 58 AM

After
Screen Shot 2022-02-24 at 6 39 11 AM

---
Agreed

Returning an error from the handler or using the default (in this case it would be the fastify-sensible) error handler would be better solutions imo, this is just meant to illustrate the problem which is why I marked do not merge. The custom error handler seems to basically just be normalizing errors from requests made with got.

davidjmurphyjr:
> using the default (in this case it would be the fastify-sensible) 

Ah, good call, I glossed over that, thanks!

---

> The custom error handler seems to basically just be normalizing errors from requests made with got.

I think you are referring to this
```
if (error instanceof HTTPError || error instanceof HttpError) {
  statusCode = error.response.statusCode
} else if (error instanceof UserError) {
  statusCode = error.statusCode
}
```

Thanks for calling it out, I did not understand what was going on there on my first pass. After another look, I'm not sure it's normalizing -- it treats built-in HTTPErrors and got HttpErrors the same. I think what it's doing is: if the private API makes an HTTP request and it fails, it forwards the failed request's status code on by setting it in the private API's response. 

I think this is functionality that could be easily preserved if we wanted to move forward with the  -- reply.send(error) instead of reply.send(object) approach.

---

What do you see as a next step for this PR/Issue, or is this purely an FYI?

davidjmurphyjr:
@afraser I'm not sure this is directly related to the "email gets sent despite throwing an error" issue, but the way I'd go about fixing that issue would depend of transaction rollbacks happening when we throw... so kinda related? We should talk again, I've been through the effected codepath in a bit more detail, I have ideas...

IanPirro:
> After another look, I'm not sure it's normalizing -- it treats built-in HTTPErrors and got HttpErrors the same. I think what it's doing is: if the private API makes an HTTP request and it fails, it forwards the failed request's status code on by setting it in the private API's response.

Yeah, admittedly a loose usage of normalize lol that's what I was getting at that they are treated the same

> What do you see as a next step for this PR/Issue, or is this purely an FYI?

Basically a FYI, in my project I ended up just deleting the custom error handler as I don't have a requirement to make outgoing requests yet via got yet, though I'll probably just make an error type that takes a got `RequestError` and throw that in the catch{} from outgoing calls. I can take a swing at it here if you want, otherwise should be pretty straightforward if you know what you want to do as far as preserving the status code.

davidjmurphyjr:
I took a stab at a fix in #302, I did a little testing and it looks pretty good to me, LMK what you think

IanPirro:
Closing in favor of #302

PR #300, opened 2022-02-24
Fix for deployment

dallashuggins:
### Changes
- Removing condition stating the preview image cannot be `null` if the NFT template is published. 
  - GCP was throwing the following error on deployment. It started error'ing when we deployed the changes to prevent null values in the snapshot config. It seems this is causing problems, so removing this conditional. 

#### Error
```
"14:14:32 :rotating_light:Failed to create field "nft_templates.preview_image""
and then
14:14:32 :rotating_light: Undefined binding(s) detected when compiling FIRST. Undefined column(s): [field] query: select "id" from "directus_fields" where "collection" = ? and "field" = ? limit ? "
```

### DIFF
--- a/apps/cms/snapshot.yml
+++ b/apps/cms/snapshot.yml
@@ -2458,15 +2458,6 @@ fields:
       foreign_key_schema: public
       foreign_key_table: directus_files
       foreign_key_column: id
-      conditions:
-        - name: No editing after publish
-          rule:
-            _and:
-              - status:
-                  _eq: published
-              - id:
-                  _nnull: true
-          readonly: true
     meta:
       collection: nft_templates
       field: preview_image


davidjmurphyjr (approved):
LGTM. 

Whaat is the plan going forward, I would guess that re-seeding would fix it, is that an option (assuming yo  are seeing this error in the demo instance)? Or, do we have to preserve the existing data, in which case do we patch the data and reintroduce the rule?

afraser (approved):
Echoing David's question here. I'm cool with merging this to fix current issues (esp. if it's blocking anyone) but I'm not sure we should stick with this and I would think reinitializing and seeding the dbs would do the trick (although we might consider trying an export/reimport strategy).

dallashuggins:
> LGTM.
> 
> Whaat is the plan going forward, I would guess that re-seeding would fix it, is that an option (assuming yo are seeing this error in the demo instance)? Or, do we have to preserve the existing data, in which case do we patch the data and reintroduce the rule?

@davidjmurphyjr Great point actually, if re-seeding fixed it that'd be ideal. For some reason this didn't even occur to me. I'm going to check in Slack if everyone would be okay re-seeding the Algomart db to see if that fixed it.

dallashuggins:
Closing since we're recreating the databases, which should fix the issue.

PR #302, opened 2022-02-25
Rollback DB transaction when an error is thrown

davidjmurphyjr:
This an alternative fix for the error Ian discovered in #296 

### To test 
1. Backup two commits to **reproduce**  -- The `POST /accounts` endpoint is set up to write a new UserAccount to the DB then make an HTTP call with got that responds 401.
    1.  The POST Accounts should respond with an error
    2.  A DB record should be written to the UserAccounts table, **this is the buggy behavior**
2. Move forward one commit to **fix**
    1.  The POST Accounts should respond with an error. **Regression test:** this error should be functionally equivalent to the error response in step 1.i  (e.g. it should not be a generic 500, it should forward the status code returned by got)
    2.  A DB record should be **not** be written to the UserAccounts table, **this is the correct behavior**

Screen cap: 

https://user-images.githubusercontent.com/3670577/155746672-f00559b5-4eb6-4132-b56b-1c415329c0e6.mp4

### DIFF
--- a/apps/api/src/modules/accounts/accounts.service.ts
+++ b/apps/api/src/modules/accounts/accounts.service.ts
@@ -6,11 +6,8 @@ import {
 } from '@algomart/schemas'
 import { UpdateUserAccount } from '@algomart/schemas'
 import { Username } from '@algomart/schemas'
-import got, { Got } from 'got'
 import { Transaction } from 'objection'
-import { v4 } from 'uuid'
 
-import { Configuration } from '@/configuration'
 import AlgorandAdapter from '@/lib/algorand-adapter'
 import { AlgorandAccountModel } from '@/models/algorand-account.model'
 import { AlgorandTransactionModel } from '@/models/algorand-transaction.model'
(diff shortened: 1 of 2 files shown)


afraser:
I'm all about this.
Question though: could (how might) this impact our checkout flow? (eg: if the submitted CC or wire form is invalid, or if circle responds with an error from `createBankAccount` or any other api of theirs we use).

IanPirro:
Looks good, to @afraser point I would just check that the error messages have what the frontend would expect from this removed line `.send(error.response.body)` when it's a `got` error

IanPirro (approved):

davidjmurphyjr:
Okay, I'll run through a pack purchase and try a couple bad CVV codes? Anything else you think I should hammer on?

afraser:
I can help you test out the wire flow as well if ya wanna hop on a call on Monday. Should be p quick

afraser (approved):
Tested the wire flow against a circle error and everything seems good.

smonn (approved):

PR #314, opened 2022-03-01
Upgrade deps

smonn:


dallashuggins (approved):
Thanks for doing this! Everything is working well still it seems

jakerainis (approved):
Thanks @smonn -- Sorry these got put on the backburner.

PR #316, opened 2022-03-04
Attempted fix for snapshot

smonn:
🤞 hoping this re-generated snapshot.yml file will fix the deployment issues

Also included a fix for a Next.js warning about including `<link rel="stylesheet" />` tags in `next/head`: https://nextjs.org/docs/messages/no-stylesheets-in-head-component


smonn (commented):
Merging this right away to verify if the deployment will work

PR #324, opened 2022-03-07
Upgrade deps

smonn:


apps/web/src/templates/nft-template.tsx
```diff
@@ -91,7 +91,9 @@ export default function NFTTemplate({
 
         {/* TODO: add tabs for description, activity, listings, and offers */}
 
-        <ReleaseDescription description={collectible.body} />
+        {collectible.body ? (
+          <ReleaseDescription description={collectible.body} />
+        ) : null}
```
smonn, line 96: Guard against empty `body` prop.

dallashuggins (approved):
Everything is working the same in the app. Thanks for doing this!

PR #340, opened 2022-03-16
Replace Got with Axios

### LINKED ISSUE #326: Replace Ky and Got with Axios
## Overview

Due to a bug in Ky (and Got?), it cannot handle large JSON payloads. So for now, we'll replace their usages with Axios.

smonn:
Closes #326 

- [x] Replace Got with Axios
- [x] Replace Ky usage with Axios in Web's API Client
- [x] Ensure API Client is only used in Web server-side
- [x] Ensure other services in Web are only used client-side (to avoid Ky usage in Node.js)

**Motivation**

Got keep triggering "Error: The `onCancel` handler was attached after the promise settled."  in the API and [Ky cannot (currently) handle large JSON requests](https://github.com/sindresorhus/ky/issues/432) in Node.js <17 environments due to its clone usage.


dallashuggins (approved):
Working great locally!

dallashuggins (commented):
I noticed once thing actually with the closing of the connection. The database connection says it closes in the logs, but it doesn't seem to ever exit.

smonn:
> I noticed once thing actually with the closing of the connection. The database connection says it closes in the logs, but it doesn't seem to ever exit.

It's a bit unfortunate timing I'm afraid. It does close the connection etc, but the logs happen _after_ your terminal prints the prompt. If you hit `[enter]` one time after seeing the logs you'll notice this (as opposed to hitting `[ctrl] - c`).

PR #356, opened 2022-03-17
Update smart contract to support royalty fee

smonn:
Auction contract is now updated to support a royalty fee paid to the creator account. Also included are changes to _only_ pay the bid amount to the seller and return everything else to the creator (including the royalty fee). One notable  difference is that when closing the auction, both the seller and creator addresses will need to be referenced in the app no-op transaction.

The current version of the contract does still _not_ allow closing it while the auction is active. This might be changed based on our needs.

When reviewing this PR, feel free to skip the `auction_aproval.teal` file as it's generated by the PyTEAL contract.

See previous PR notes below.

---

Some early draft notes:

```
actors:
  banker (funding wallet owned by algomart)
  creator (temporary account that creates the smart contract)
  seller (original NFT holder, custodial account on algomart)
  buyer (account that places one or more bids)
  escrow (account associated with the smart contract)

seller selects options for auction:
  nft (algorand standard asset, must be owned by the seller)
  reserve price (minimum accepted bid in ALGO)
  duration (how many hours should the auction be live for)
  minimum bid increment (optional/should this be a user option? default to 0.1 ALGO?)
  passphrase (to decrypt private key/mnemonic)

once submitted, the system needs to take these actions:
  setup creator account:
    banker sends X ALGO to creator (what does X need to be?)
    estimate: 0.1 (min balance) + 0.1 (for NFT) + 0.1 (hold smart contract) + 0.0285 * 8 (for each global state uint) + 0.05 * 2 (for each byte slice) + 0.203 (escrow funds) + 0.004 (create smart contract + setup of smart contract) = 0.3 + 0.228 + 0.1 + 0.207 = 0.835 ALGO
    excluding the transaction fees (totalling 0.007 ALGO), these funds _should_ be recoverable
  transfer nft to creator account:
    this is to ensure we can take action without user intervention
    from a UI perspective, the NFT is still owned by the seller
  create smart contract:
    creator creates the smart contract with params based on seller options
  configure smart contract:
    creator funds the escrow account with 0.203 ALGO
    creator configures the escrow account (select which NFT is for sale)
    creator transfers NFT to escrow account
```

### DIFF
--- a/contracts/src/contracts/auction.py
+++ b/contracts/src/contracts/auction.py
@@ -13,6 +13,14 @@ def approval_program():
     num_bids_key = Bytes("num_bids")
     lead_bid_amount_key = Bytes("bid_amount")
     lead_bid_account_key = Bytes("bid_account")
+    # 0.1 ALGO (min account balance)
+    # 0.1 ALGO (holding asset)
+    # 4 x 1000 microAlgo (transaction fees, see below)
+    # 1. opt in to asset
+    # 2. transfer asset to winner
+    # 3. transfer bid amount to seller
+    # 4. transfer royalty to creator
+    escrow_min_balance = Int(204_000)
 
     @Subroutine(TealType.none)
     def closeNFTTo(assetID: Expr, account: Expr) -> Expr:
@@ -114,8 +122,37 @@ def approval_program():
         Approve(),
     )
 
+    on_setup_payment_txn_index = Txn.group_index() - Int(1)
+    on_setup_asset_txn_index = Txn.group_index() + Int(1)
     on_setup = Seq(
-        Assert(Global.latest_timestamp() < App.globalGet(start_time_key)),
+        Assert(
+            And(
+                Global.latest_timestamp() < App.globalGet(start_time_key),
+                # assert previous txn is payment of min balance
+                Gtxn[on_setup_payment_txn_index].type_enum() == TxnType.Payment,
+                Or(
+                    Gtxn[on_setup_payment_txn_index].sender()
+                    == Global.creator_address(),
+                    Gtxn[on_setup_payment_txn_index].sender()
+                    == App.globalGet(seller_key),
+                ),
+                Gtxn[on_setup_payment_txn_index].receiver()
+                == Global.current_application_address(),
+                Gtxn[on_setup_payment_txn_index].amount() == escrow_min_balance,
+                # assert next txn is transfer of NFT
+                Gtxn[on_setup_asset_txn_index].type_enum() == TxnType.AssetTransfer,
+                Or(
+                    Gtxn[on_setup_asset_txn_index].sender() == Global.creator_address(),
+                    Gtxn[on_setup_asset_txn_index].sender()
+                    == App.globalGet(seller_key),
+                ),
+                Gtxn[on_setup_asset_txn_index].asset_receiver()
+                == Global.current_application_address(),
+                Gtxn[on_setup_asset_txn_index].asset_amount() >= Int(1),
+                Gtxn[on_setup_asset_txn_index].xfer_asset()
+                == App.globalGet(nft_id_key),
+            )
+        ),
         # opt into NFT asset -- because you can't opt in if you're already opted in, this is what
         # we'll use to make sure the contract has been set up
         InnerTxnBuilder.Begin(),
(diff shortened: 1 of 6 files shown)


contracts/src/contracts/auction.py
```diff
@@ -64,6 +65,29 @@ def closeAccountTo(account: Expr) -> Expr:
...
+                    TxnField.receiver: App.globalGet(seller_key),
+                    TxnField.amount: (
+                        App.globalGet(lead_bid_amount_key)
+                        - (
+                            (
+                                App.globalGet(fee_percent_key)
+                                * App.globalGet(lead_bid_amount_key)
+                            )
+                            / Int(100)
+                        )
```
smonn, line 84: Integer math is fun! This is the same as `bidAmount * (1 - (royaltyFee / 100))` where `royaltyFee` is the percentage express as a whole number from 0 to 100 (inclusive).

contracts/src/contracts/utils.py
```diff
@@ -534,10 +542,14 @@ def closeAuction(client: AlgodClient, appID: int, closer: Account):
...
     appGlobalState = getAppGlobalState(client, appID)
+    appCreator = getAppCreator(client, appID)
 
     nftID = appGlobalState[b"nft_id"]
 
-    accounts: List[str] = [encoding.encode_address(appGlobalState[b"seller"])]
+    accounts: List[str] = [
+        encoding.encode_address(appGlobalState[b"seller"]),
+        appCreator,
+    ]
```
smonn, line 552: Application creator address is required to be able to send the royalty fee to that account.

contracts/src/contracts/utils.py
```diff
@@ -555,3 +567,28 @@ def closeAuction(client: AlgodClient, appID: int, closer: Account):
     client.send_transaction(signedDeleteTxn)
 
     waitForTransaction(client, signedDeleteTxn.get_txid())
+
+
+def fundAccount(client: AlgodClient, funder: Account, recipient: Account, amount: int):
```
smonn, line 572: Not currently used. Added this to support a scenario more closely matching how we will do things in AlgoMart's Node.js code (funding wallet sends funds to generated creator account, then creator account creates the smart contract). Leaving it in here in case we decide to use it.

contracts/tests/test_auction.py
```diff
@@ -18,18 +20,30 @@
     setupAuctionApp,
 )
 
+# min balance: 0.1 ALGO
+# smart contract, base: 0.1 ALGO
+# smart contract, bytes: 0.05 * 2 = 0.1 ALGO
+# smart contract, ints: 0.0285 * 8 = 0.228 ALGO
+# smart contract, escrow: 0.1 + 0.1 + 0.003 = 0.203 ALGO (min, nft, txns)
+# smart contract, create + setup txns: 0.001 * 2 = 0.002 ALGO
+CREATOR_FUNDS = 100_000 + 100_000 + 100_000 + 228_000 + 203_000 + 2000
```
smonn, line 29: See my comment about `fundAccount`. Also an approximate for how much is needed by the creator account.

contracts/src/contracts/utils.py
```diff
@@ -435,7 +443,7 @@ def setupAuctionApp(
         # additional min balance to opt into NFT
         + 100_000
         # 3 * min txn fee
-        + 3 * 1_000
+        + 4 * 1_000
```
jakerainis, line 446: Was this changed to 4 for testing, or is intentional? If so, I presume the comment above should change to 4 as well?

davidjmurphyjr (approved):
Looks good!

contracts/src/contracts/utils.py
```diff
@@ -191,6 +197,13 @@ def getLastBlockTimestamp(client: AlgodClient) -> Tuple[int, int]:
     return block, timestamp
 
 
+def waitUntilTimestamp(client: AlgodClient, minTimestamp: int):
```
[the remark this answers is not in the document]
smonn: It's used only in `test_auction.py` to replace plain `sleep`. I made the change while trying to figure out why the tests were taking so long. It turns out, the issue was with my local Algorand sandbox being off by one hour. Most likely due to the recent DST change. I kept it in as it was more explicit compared to the previous code with what it does.

PR #357, opened 2022-03-18
Refactor shared utils into lib

eremzeit:
This is a foundational PR I made for the future work of creating the scribe service.  I limited the scope of this PR to include:
- moving shared utility functions from the `api` app to a new nx library at `libs/shared/utils`

but to support those changes I needed to also...
- refactor some utility functions to remove the usage of any global configuration variables
- notable: Because of a limitation in how tsconfig file inheritence works, we can only define our module mappings either at the root config file or the leaves.  The recommended way in Nx is to put them in the root and indeed the nx code generators won't work if you put them in the leaves.  A consequence is that I have to replace any instance of the `@/` path mapping with the new alias that I created `@api/`.


.eslintrc.js
```diff
@@ -37,6 +37,13 @@ const defaultRules = {
   // Allow using any for rest arguments (i.e. `...any[]`)
   '@typescript-eslint/no-explicit-any': ['warn', { ignoreRestArgs: true }],
 
+  '@nrwl/nx/enforce-module-boundaries': [
+    2,
+    {
+      allowCircularSelfDependency: true,
+    },
+  ],
```
smonn, line 45: Is this a temporary fix while other refactoring is being done? Or has it more to do with a "false positive" due to the `@/` -> `@api/` change? I just noticed Nx does recommend using relative paths within an app/library and using `@project/lib` style imports for library imports.

README.md
```diff
@@ -134,17 +134,27 @@ When creating a new package, first determine which kind of package you are creat
...
-Shared packages used by this monorepo.
+For performance and code organization reasons, the [https://nx.dev/structure/applications-and-libraries](Nx docs) recommend
+putting as much functionality as possible into libs, even if the code is only used in a single app. In Nx, a lib is more than just a directory under the `libs/` directory. Each lib must have an entry in the workspace.json file for the lib to build and import correctly.
 
-- [`schemas`][schemas] - Shared code and typings
+Linting will fail for any lib code that tries to import code from an app. This means that lib code should never access things like
+global configuration variables or environment variables. (eg. `Configuration`)
+
+If you wanted to create a new library at the path `libs/shared/utils`, you'd use the nx generator...
+`nx generate @nrwl/node:lib utils --directory shared`
```
smonn, line 152: Is the intention to put **all** shared code in `@algomart/shared`? Any reason not to do multiple shared libraries such as `@algomart/algorand`, `@algomart/utilities`, `@algomart/payments` etc? Because it looks like the import paths will need to be distinct per "sub-library" anyway:

```json
  // from tsconfig.base.json
  "@algomart/shared/utils": ["libs/shared/utils/src/index.ts"],
```

apps/api/src/modules/packs/packs.routes.ts
```diff
@@ -24,9 +24,13 @@ export async function getPublishedPacks(
   }>,
   reply: FastifyReply
 ) {
+  console.log('IN HANDLER')
```
smonn, line 27: These for debugging purposes? Can they be removed?

smonn (commented):
Thanks for working through all of this! Left some comments in here. Happy to jump on a call if you want to discuss.

README.md
tj-boyle: Screen Shot 2022-03-22 at 1 38 43 PM

This is what it looks like in the client's implementation - I agree we can probably move a number of these out of the somewhat redundant "shared" folder

eremzeit: I mimicked the directory structure recommended in the Nx docs.  It's also how nx example project is setup.
https://github.com/nrwl/nx-examples/tree/master/libs/shared

image

image

.eslintrc.js
eremzeit: The linting rule source code documents the `allowCircularSelfDependency` as follows: `Projects should use relative imports to import from other files within the same project. Use "./path/to/file" instead of import from "{{imp}}"`

I had to relax this rule because I was getting linting rule failures for existing code that I hadn't written complaining about using module aliases to reference the app that you are currently in.  For example, if you had a module in the api app that had an import that used `@api/[...]`, the linter would complain and tell you to replace the alias with a relative path.

I don't have an opinion about relative paths versus aliases.  Actually when I made these changes before on "the client's" fork of this repo, I had originally just fixed all the linting errors.  But then in that PR people said they liked having the ability to use `@api/` within the api project so I had to revert the changes and relax the linting rule.  That took me quite a while to do so this time I just left it be.

README.md
eremzeit: An example of something that would go in `libs/api` would be endpoint handlers that only live in api.... eg. the code that currently lives in `apps/api/modules`.

eremzeit: @smonn I just reread your comment.  To clarify, the directory `libs/shared` isn't its own library.  If you look in workspaces.json you can see that each individual sub-directory is it's own library, which means that they can each get built and execute tests in parallel.

smonn: That all makes sense. Happy to lean into the recommended Nx approach here.

smonn (approved):

PR #358, opened 2022-03-18
Move models to shared lib

eremzeit:
(this is based off of `aj/refactor-shared-utils` so that should be merged first)

In this PR I move all of our models into a shared lib at `libs/shared/models`


smonn (approved):
👍

PR #366, opened 2022-03-21
Tj/currency language

tj-boyle:
This PR:
- Adds currency dropdown
  - In footer
  - In registration
  - In edit profile
- Add language dropdown
  - In footer
  - In registration
  - In edit profile
- Changing currency converts values across app
- Separates language from locale
  - Language used for static text translations, grabbed from dropdown selection (with a series of cascading defaults)
  - Locale used for formatting date, currency


apps/api/src/migrations/20220324201352_add_lang_currency.ts
```diff
@@ -0,0 +1,35 @@
...
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable('UserAccount', (table) => {
+    table.renameColumn('locale', 'language')
+    table.string('currency').notNullable().defaultTo('USD')
+  })
+
+  await knex.schema.createTable('CmsCacheLanguages', (table) => {
+    table.string('code').primary()
+    table.string('name').notNullable()
+    table.string('sort').unique().notNullable()
```
jakerainis, line 12: I think we want to allow this to be nullable since sort values aren't assigned by default. I ran into an issue locall running this but was able to fix it by making this nullable and changing the sort field for languages in the CMS to "Allow NULL" in the field settings.

jakerainis (approved):
@tj-boyle Left one comment re: constraints but looks great. Might be worth getting this one in and then we can rebase the scribe PRs with the relevant services and migrations.

apps/api/src/migrations/20220324201352_add_lang_currency.ts
tj-boyle: I opted to add a default to of 1, thanks for the callout!

PR #367, opened 2022-03-22
Aj/refactor various shared code 

eremzeit:
In this PR I refactor `adapters`, `plugins`, and `services` out of api and into a shared lib.


eremzeit:
Heads up, I just noticed that I still need to do a few minor fixes to imports in the task files... so this PR is technically WIP.

eremzeit:
Okay apparently after I merged my previous PR it caused some merge conflicts, but I have a feeling these aren't "real" and that when I rebase using the git CLI it will automatically handle some of these.

libs/shared/services/src/lib/packs.service.ts
```diff
@@ -1334,7 +1337,7 @@ export default class PacksService {
               variables: {
                 amount: `${formatIntToFloat(
                   selectedBid.amount,
-                  Configuration.currency // TODO: receive as argument
+                  this.currency // TODO: receive as argument
```
smonn, line 1340: TODO comment can be removed, I think?

```suggestion
                  this.currency
```

smonn (approved):
Works on my end. Nicely done!

PR #370, opened 2022-03-22
CMS Cache

tj-boyle:
Adds
- New tables for CMS cache
- Task to sync CMS items 
- Adapter for CMS cache
- Swap out all non-sync calls to Directus with calls to CMS Cache


jprincipe (approved):
it's beautiful

afraser:
Very curious to hear more about this. Are you solving a performance problem here or is this just to make things generally faster? In either case I'd love to see performance comparisons (Simon used `ab` on the command line for SPIN but also pointed out https://github.com/alexfernandez/loadtest).

Cache performance wins are definitely exciting but they also make me _super_ nervous because cache-related bugs are the worst. Just want to make sure we need this.

apps/api/src/tasks/sync-cms-cache.task.ts
```diff
@@ -0,0 +1,27 @@
+import { DirectusAdapter } from '@algomart/shared/adapters'
+import { DependencyResolver } from '@algomart/shared/utils'
+import { logger } from '@api/configuration/logger'
+
+export default async function syncCMSCacheTask(registry: DependencyResolver) {
```
afraser, line 5: What's this for exactly? I'd think that anything that gets out of sync would be invalidated immediately. Is that wrong?

jprincipe:
> Very curious to hear more about this. Are you solving a performance problem here or is this just to make things generally faster? In either case I'd love to see performance comparisons (Simon used `ab` on the command line for SPIN but also pointed out https://github.com/alexfernandez/loadtest).
> 
> Cache performance wins are definitely exciting but they also make me _super_ nervous because cache-related bugs are the worst. Just want to make sure we need this.

This is a critical change to support scaling the app past a few instances.  Without this change you need to scale the CMS at the same rate as the API.  

We weren't able to get to performance testing with the original client this was written for, but I would expect a significant performance gain.  Above and beyond the scaling updates, it removes a number of N+1 round trips to the CMS, and in general removes a lot of overhead of an additional network round trip.

As for reliability on the cache it is a real time update of data.  The application leverages web hooks from Directus, and if anything fails it will alert on error.

tj-boyle:
@afraser

In essence this moves the Directus DB information over to the main Algomart DB, and removes making calls to Directus in favor of doing queries against the Algomart DB.

The benefits being as Jon outlined

apps/api/src/tasks/sync-cms-cache.task.ts
jprincipe: this is run on the initial load of the application.  Ideally it only runs on application start, rather than hourly

libs/shared/services/src/lib/packs.service.ts
```diff
@@ -281,31 +271,20 @@ export class PacksService {
...
-      sortBy,
-      sortDirection,
-    })
-
     const assemblePack = this.createPublishedPackFn(
       packLookup,
       packWithActiveBidsLookup
     )
 
+    const postQueryFilters = this.createPackFilterFn({ status, reserveMet })
```
tj-boyle, line 279: `status` is a derived field populated in `toPackBase` by comparing `releasedAt` and `auctionUntil` dates.

I could make a query that makes filters following the logic in `toStatus` (within CMSCacheAdapter), but that'd involve me adding some filter complexity around handling `orWhere` queries vs `where` queries (all of the status queries would need to be `or` conditionals, and right now the query builder is all `and` conditionals)

Additionally, I'd be duplicating the logic in toStatus, which could have odd effects if the two behave slightly differently.

Ultimately I opted to just filter it post DB lookup for now because of these considerations

tj-boyle: `reserveMet` is comparing the `activeBid` against the pack price. `activeBid` is derived in the `getPacksWithActiveBids` mapping step of this function, it's not stored on the db. 

Again this is a very timely value so I can't cache the field in the DB.

We also don't know what the active bid is until doing subsequent lookups after querying the cache table

Thus, it too lives as a post DB lookup

apps/api/src/tasks/sync-cms-cache.task.ts
afraser: Ok yeah, we should probably change that then.

jakerainis (approved):
Wahoo! Thanks for porting this over. It makes local development so much easier, I'm looking forward to working with it in the future.

apps/cms/scripts/seed.mjs
```diff
@@ -127,6 +127,13 @@ async function main(args) {
     languages
     CASCADE`)
 
+  await knex.raw(
+    `INSERT INTO directus_webhooks (
+      "name", "method", "url", "status", "data", "actions", "collections", "headers"
+    ) VALUES (
+      'CMS Cache Content', 'POST', 'http://localhost:3002/webhooks/directus', 'active', 't', 'create,update,delete', 'application,collections,countries,faqs,homepage,languages,nft_templates,pack_templates,page,rarities,sets', NULL
+    )`
+  )
```
smonn, line 136: This definitely works for now, I think? Won't this try to hit an endpoint that doesn't exist until Scribe is merged in?

I guess a nice-to-have would be to instead use the Directus API during Scribe's startup to ensure a webhook is configured: https://docs.directus.io/reference/system/webhooks/

tj-boyle: Yeah I'll pull this out and we can have scribe set it up

apps/api/src/seeds/seed-test-data.ts
```diff
@@ -53,7 +55,7 @@ export const algorandAccountFactory = Factory.define<AlgorandAccount>(
     (creationTransaction) => creationTransaction.id
   )
   .attr('encryptedKey', ['mnemonic', 'passphrase'], (mnemonic, passphrase) =>
-    encrypt(mnemonic, passphrase, TEST_APP_SECRET)
```
eremzeit, line 56: Good catch.  I was trying to simplify certain things by introducing `TEST_APP_SECRET`, but seeing it in someone elses PR I can see how that would cause problems.

eremzeit: @tj-boyle Just a heads up, eventually this file might be shared between api and scribe codebases, which means it will need to be moved out of the api app, which means that we'll have to refactor out the use of `Configuration.secret`.  When/if that eventually happens I guess we'd just modify the factory to accept the secret as a parameter.

tj-boyle: yup! good callout

apps/api/src/tasks/sync-cms-cache.task.ts
eremzeit: @jprincipe ^

libs/shared/services/src/lib/packs.service.ts
tj-boyle: Okay I did a DB lookup after banging my head against a wall for a day and a half, because otherwise we'd be filtering down the singular pages results which isn't good

apps/api/src/tasks/sync-cms-cache.task.ts
tj-boyle: Ideally if any of them failed the task would fail. Idt any of them need to be run in order, but IMO I'd rather just know the sync was failing for some reason asap and have it prevent the rest because without them all there's going to be problems lurking

smonn (approved):

PR #371, opened 2022-03-23
Wire amount formatting; winning bidder fix; media gallery fixes

lefnire:
`git cherry-pick`'d a bucket-list of items. Need to verify some of the image stuff in the morning, seems some commit competition that's direction rather than syntax; so I'll check in on that & ping here when GTG.


lefnire:
I think GTG. @afraser wanna take a look? BTW merge blocked pulling from lefnire/algomart, I don't have perms to this repo to create branch here

PR #374, opened 2022-03-23
Wire amount formatting; winning bidder fix; media gallery fixes

lefnire:
`git cherry-pick`'d a bucket-list of items. Need to verify some of the image stuff in the morning, seems some commit competition that's direction rather than syntax; so I'll check in on that & ping here when GTG.

New PR from `origin` instead of `lefnire` (https://github.com/deptagency/algomart/pull/371)

### DIFF
--- a/apps/web/languages/en-US/common.json
+++ b/apps/web/languages/en-US/common.json
@@ -60,7 +60,8 @@
     "View My Collection": "View My Collection",
     "View On AlgoExplorer": "View on AlgoExplorer",
     "View Release": "View Release",
-    "View Set": "View Set"
+    "View Set": "View Set",
+    "View image X of Y":"View image X of Y"
   },
   "dateTime": {
     "Days_one": "Day",
@@ -169,6 +170,7 @@
     "Collected": "Collected",
     "Checking for Payment": "Checking for Payment",
     "Connected to Wallet": "Connected to Wallet",
+    "Cover Image": "Cover Image",
     "Creating Payment": "Creating Payment",
     "Free": "Free",
     "I don’t have a code": "I don’t have a code",
--- a/apps/web/languages/en-US/common.json
+++ b/apps/web/languages/en-US/common.json
@@ -61,7 +61,7 @@
     "View On AlgoExplorer": "View on AlgoExplorer",
     "View Release": "View Release",
     "View Set": "View Set",
-    "View image X of Y":"View image X of Y"
+    "View image X of Y":"View image {{x}} of {{y}}"
   },
   "dateTime": {
     "Days_one": "Day",
--- a/apps/web/src/components/featured-pack/featured-pack.tsx
+++ b/apps/web/src/components/featured-pack/featured-pack.tsx
@@ -73,7 +73,7 @@ export default function HomeTemplate({
         {/* Image */}
         <div className={css.featuredImage}>
           <Image
-            src={featuredPack.image}
+            src={`${featuredPack.image}?width=1280&quality=75`}
             width={512}
             height={512}
             layout="responsive"
--- a/apps/web/src/components/featured-pack/featured-pack.tsx
+++ b/apps/web/src/components/featured-pack/featured-pack.tsx
@@ -73,12 +73,13 @@ export default function HomeTemplate({
         {/* Image */}
         <div className={css.featuredImage}>
           <Image
-            src={`${featuredPack.image}?width=1280&quality=75`}
+            src={featuredPack.image}
             width={512}
             height={512}
+            sizes="(max-width: 639px) 100vw, (max-width: 767px) calc(100vw * 1/2), (min-width: 768px) calc(100vw * 3/5)"
             layout="responsive"
             objectFit="cover"
-            alt={featuredPack.title}
+            alt={t('common:statuses.Cover Image')}
             priority
           />
         </div>
--- a/apps/web/src/components/media-gallery/media-gallery.tsx
+++ b/apps/web/src/components/media-gallery/media-gallery.tsx
@@ -76,7 +76,7 @@ export default function MediaGallery({ media }: MediaGalleryProps) {
               alt={t('common:statuses.Selected Image')}
               className={css.contain}
               layout="fill"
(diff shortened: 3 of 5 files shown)


apps/web/src/components/featured-pack/featured-pack.tsx
```diff
@@ -73,7 +73,7 @@ export default function HomeTemplate({
         {/* Image */}
         <div className={css.featuredImage}>
           <Image
-            src={featuredPack.image}
+            src={`${featuredPack.image}?width=1280&quality=75`}
```
afraser, line 76: So we were using this when we were loading images directly from the CMS which we aren't anymore (in my other PR this was just an `img` tag. We're now using `Image` tag without the cms image loader). What we need here now is actually a `sizes` prop configured to specify widths at various breakpoints.

Here's what this looks like in SPIN now:
https://github.com/rocketinsights/algomart-spin/blob/8d5789779fbfd798026de25da55e04a6568b4a61/apps/web/src/components/featured-pack/featured-pack.tsx#L82

Unfortunately I'm not 100% sure the widths we use on SPIN are exactly the same as what we use here. Luckily, you don't have to specify the widths in pixel terms though so you can instead just resize the window and note whether the image should be approximately `100vw`, `50vw`, `33vw` etc, then use the same media syntax.

apps/web/src/components/media-gallery/media-gallery.tsx
```diff
@@ -76,7 +76,7 @@ export default function MediaGallery({ media }: MediaGalleryProps) {
               alt={t('common:statuses.Selected Image')}
               className={css.contain}
               layout="fill"
-              src={currentMedia}
+              src={`${currentMedia}?width=700&quality=95`}
```
afraser, line 79: Simon already set a `sizes` prop on this one so there's no need for any changes here.

apps/web/src/components/media-gallery/media-gallery.tsx
```diff
@@ -102,10 +102,10 @@ export default function MediaGallery({ media }: MediaGalleryProps) {
                 onClick={() => handleMediaChange(media)}
               >
                 <Image
-                  alt=""
+                  alt="media"
                   className={clsx(css.cover, css.fullWidth, css.rounded)}
                   layout="fill"
-                  src={media}
+                  src={`${media}?width=300`}
```
afraser, line 108: this one still needs `sizes`. I _believe_ we use a 4-column grid for all screen sizes here so it might be as simple as 
 `sizes="25vw"`

afraser (changes requested):

apps/web/src/components/media-gallery/media-gallery.tsx
smonn:
```suggestion
                  src={media}
                  sizes="25vw"
```

apps/web/src/components/media-gallery/media-gallery.tsx
```diff
@@ -102,10 +102,10 @@ export default function MediaGallery({ media }: MediaGalleryProps) {
                 onClick={() => handleMediaChange(media)}
               >
                 <Image
-                  alt=""
+                  alt="media"
```
smonn, line 105: Let's add a translation for the `alt` attribute, something like this maybe?

```suggestion
                  alt={t('common:actions.View image X of Y')}
```

smonn (dismissed):
Some minor things, looks good overall!

apps/web/src/components/featured-pack/featured-pack.tsx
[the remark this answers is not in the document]
lefnire: Addressed the comments here, give the latest a gander (force-pushed due to origin/main rebase). Especially check out this file to see if I did it right, I think I did (did some visual testing)

apps/web/languages/en-US/common.json
```diff
@@ -60,7 +60,8 @@
     "View My Collection": "View My Collection",
     "View On AlgoExplorer": "View on AlgoExplorer",
     "View Release": "View Release",
-    "View Set": "View Set"
+    "View Set": "View Set",
+    "View image X of Y":"View image X of Y"
```
smonn, line 64: You'll wanna make this a template translation, like so:

```suggestion
    "View image X of Y":"View image {{x}} of {{y}}"
```

apps/web/src/components/media-gallery/media-gallery.tsx
```diff
@@ -102,11 +102,12 @@ export default function MediaGallery({ media }: MediaGalleryProps) {
                 onClick={() => handleMediaChange(media)}
               >
                 <Image
-                  alt=""
+                  alt={t('common:actions.View image X of Y')}
```
smonn, line 105: Using a template translation (grab `i` from the map call):

```suggestion
                  alt={t('common:actions.View image X of Y', { x: i, y: media.length )}
```

smonn, line 105: Using a template translation (grab `i` from the map call):

```suggestion
                  alt={t('common:actions.View image X of Y', { x: i, y: media.length )}
```

smonn (commented):
A couple of small tweaks, thanks!

apps/web/languages/en-US/common.json
lefnire: fixed (force-push to rebase)

apps/web/src/components/media-gallery/media-gallery.tsx
jakerainis: Should this be `i+1`?

lefnire: I pushed as i+1

afraser:
This PR is getting a bit crusty. If I'm not mistaken it just needs to be updated from main and to have one small Image update that Simon suggested.

lefnire:
@afraser I think I've got the simon points squared, and I just rebased to origin/main

afraser (approved):

afraser:
Thanks @lefnire !

PR #378, opened 2022-03-28
Creating scribe app.  Moving tasks.

eremzeit:
# Overview 
- created the scribe app, which currently sets up a fastify instance with sane defaults 
- Refactored tasks out of api into a library
- Moved the execution of the tasks out of the api app and into the scribe service 

## Changelog notes
- ***important: Once this PR is merged, running the scribe service will be required in order to get tasks to run.  In a dev environment you can just use `npm run start` like normal, but in a deployed environment that probably means setting up an entirely new running service (default port 3002)
- Because the tasks are now executed in the scribe app, you'll need to create a `.env` file for the scribe app.  I have included an `.env.example`.  
- Most importantly, you'll need make sure tha the `ENABLE_JOBS` is set to true on the scribe app rather than the API app which no longer handles job execution.


apps/scribe/src/configuration/configure-resolver.ts
```diff
@@ -0,0 +1,223 @@
...
+  resolver.set(
+    BidsService.name,
+    (c) =>
+      new BidsService(
+        c.get<NotificationsService>(NotificationsService.name),
+        c.get<PacksService>(PacksService.name),
+        Configuration.currency,
+        logger
+      )
+  )
```
dallashuggins, line 105: Currently seeing an error about the amount of arguments when I try to start up the app:

Screen Shot 2022-04-05 at 10 57 47 AM

apps/scribe/src/configuration/configure-resolver.ts
```diff
@@ -0,0 +1,223 @@
...
+    (c) =>
+      new PacksService(
+        c.get<CMSCacheAdapter>(CMSCacheAdapter.name),
+        c.get<CollectiblesService>(CollectiblesService.name),
+        c.get<NotificationsService>(NotificationsService.name),
+        c.get<AccountsService>(AccountsService.name),
+        Configuration.currency,
+        logger
+      )
+  )
```
dallashuggins, line 137: The other arguments error was for this line

PR #390, opened 2022-04-01
Minor style updates

afraser:
- Clicking "view collection" doesn't do anything when you're already on the "my collectibles" page
- match button styles to designs
- fix contrast ratio of the "New" badge in my collection
- LinkButton should be inline-block (not inline)
- Add hover state to Read More link
- fix default border color so we mostly no longer need to apply base-border (and remove base-border references in a handful of places)
- remove shadows completely from dark mode (produces ugly "banding")

### DIFF
--- a/apps/web/src/utils/urls.ts
+++ b/apps/web/src/utils/urls.ts
@@ -19,11 +19,7 @@ export const urls = {
   mySet: '/my/sets/:setSlug',
   myShowcase: '/my/showcase',
   nft: '/nft/:assetId',
-  nftDetails: '/nft/:assetId/details',
-  nftActivity: '/nft/:assetId/activity',
-  nftSell: '/nft/:assetId/sell',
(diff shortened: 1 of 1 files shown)


apps/web/tailwind.config.js
```diff
@@ -34,8 +37,8 @@ module.exports = {
         20: '20px',
       },
       boxShadow: {
-        medium: '0px 1px 10px rgba(0, 0, 0, 0.15)',
-        large: '0px 5px 40px rgba(0, 0, 0, 0.15)',
+        medium: 'var(--shadowMedium)',
```
IanPirro, line 40: 💯

IanPirro (approved):

apps/web/tailwind.config.js
afraser: still need to reconcile these with `shadow-md` and `shadow-lg` which are _also_ used throughout the app and are quite different from these guys.

PR #391, opened 2022-04-02
Various fixes...

smonn:
- Load public/browser config asynchronously
- Remove public runtime config
- Make homepage use ISR to speed up rendering
- Add useConfig, useFirebaseApp and useAnalytics hooks
- Add /api/v1/config endpoint
- Fixes to docker-compose and dockerfiles (use bullseye instead of alpine)
- Ensure sharp is available at runtime for Next.js
- Fix (hopefully) for CMS snapshot
- Tweaks to logger config and only use pino-pretty the recommended way
- Fix seed script to use correct auth
- Rebuild Directus extensions


apps/api/src/configuration/index.ts
```diff
@@ -87,35 +71,24 @@ export const Configuration = {
...
+    return env.get('CMS_ACCESS_TOKEN').required().asString()
   },
 
   get webUrl() {
     return env.get('WEB_URL').default('http://localhost:3000').asUrlString()
   },
 
   get pinataApiKey() {
-    return env.get('PINATA_API_KEY').default('').asString()
+    return env.get('PINATA_API_KEY').required().asString()
```
afraser, line 83: ah this probably makes debugging missing env vars a lot easier eh? 🙏🏼

apps/api/src/api/build-app.ts
```diff
@@ -98,5 +98,9 @@ export default async function buildApp(config: AppConfig) {
   await app.register(paymentRoutes, { prefix: '/payments' })
   await app.register(setsRoutes, { prefix: '/sets' })
 
+  app.all('/', (_, reply) => {
+    reply.send('ok')
+  })
```
smonn, line 103: This was added so that the wait-on scripts will work correctly.

apps/api/src/configuration/index.ts
smonn: 100%

Though it makes running the build/test workflows a tad more annoying.

dallashuggins (approved):
Everything has been working well running the app locally! Reset the databases and everything started up as expected, and was able to checkout, etc. I didn't try Docker though, as I haven't been using. Looks good!

PR #399, opened 2022-04-05
Apply URL Search params to releases page

jakerainis:
Filters are now conjoined with the URL for deep linking (and also helps with load testing).

### DIFF
--- a/apps/web/src/pages/releases/index.tsx
+++ b/apps/web/src/pages/releases/index.tsx
@@ -3,7 +3,7 @@ import { GetServerSideProps } from 'next'
 import { useRouter } from 'next/router'
 import useTranslation from 'next-translate/useTranslation'
 import { parse, stringify } from 'query-string'
-import { useEffect, useMemo, useRef } from 'react'
+import { useEffect, useMemo } from 'react'
 
 import { ApiClient } from '@/clients/api-client'
 import { PackFilterProvider } from '@/contexts/pack-filter-context'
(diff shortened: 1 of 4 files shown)


apps/web/src/pages/releases/index.tsx
```diff
@@ -23,7 +26,28 @@ export default function Releases({ packs }: PublishedPacks) {
...
+  const initialSelectOptions = getSelectSortingOptions(t)
+  const selectedOption = getSelectSortingOptions(t).find(
+    (option) => option.id === initialState.selectedOption
+  )
+
+  // Set initial filter state based off of URL parsing
+  const { dispatch, state } = usePackFilter({
+    ...initialState,
+    selectedOption: selectedOption ? selectedOption : initialSelectOptions[0],
+  })
```
smonn, line 49: Should some/most of this rely on `useMemo` to avoid recalculations/parsing each render?

afraser (changes requested):
Hey Jake, as I update price high/low the url query doesn't update until I toggle something else. I just noticed yesterday that we have a number of inputs on the site that are uncontrolled so my guess is that's what's going on.

afraser (approved):
spoke too soon, totally spaced on the "apply" button. Lol, I'm too used to the "apple way".

apps/web/src/pages/releases/index.tsx
afraser: We could memoize the `find` but it's searching through 2 options so it feels like overkill.

smonn: More thinking about the initial stringify/parse step. Also might help with avoiding the useEffect further down.

smonn:
One thing I noticed, which I don't know is new with this PR or happened before too, but it "bounces up" to the top of the page and then scrolls back down for each filter change. Is there a way to avoid that?

smonn:
> @jakerainis while you're in here, would you mind updating the numeric spacers in use-pack-filter?
> 
> ```
> priceHigh: 50_000
> ```
> 
> ☝🏼 this is misleading since this includes cents

@afraser That's 50,000 in whatever unit is the current one, which might be USD, SEK, or whatever the user configured. In addition, Dinero will handle the formatting of this, which in turn includes a currency definition and how many decimals should be used (i.e. what the exponent value is). I think it's fine to leave this as is.

apps/web/src/pages/releases/index.tsx
jakerainis: I added memoization. The `useEffect` was necessary since it has to update once the state was changed... Open to other ideas though...?

jakerainis:
Removed the scroll behavior as well.

smonn (approved):

PR #404, opened 2022-04-07
Use PaymentProvider & context

afraser:
Use `PaymentProvider` and `paymentContext` to reduce prop drilling in the checkout flows.


dallashuggins:
Great improvements! 🔥 Everything was working well when purchasing a pack with a purchase type, and errors were being displayed as expected. 

Whenever I try to make a bid on an auction, I get the following error from the `/add-to-pack` endpoint:
```
"Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:\n1. You might have mismatching versions of React and the renderer (such as React DOM)\n2. You might be breaking the Rules of Hooks\n3. You might have more than one copy of React in the same app\nSee [https://reactjs.org/link/invalid-hook-call]() for tips about how to debug and fix this problem."
```

Are you seeing that? Also, it never shows crypto and wires as payment options. When console logging inside the `checkout-template`, the same environment variables (console logged at the top of the component) show up true for the console logs in the terminal but false for the console logs in the developer tools (shown below). Are you seeing crypto/wire as options? 

Screen Shot 2022-04-08 at 11 08 23 AM
Screen Shot 2022-04-08 at 11 08 05 AM

afraser:
@dallashuggins Thanks so much for taking a look!

I haven't see the invalid hook call yet. I'll try and reproduce and fix that now. 

re: never showing crypto and wires, I noticed this but I'm actually seeing this happen in `main` too. I thought it was just my environment. It's super weird because if I log `isEnabled` and `isEnabled.toLowerCase() === 'true'` in `get isWireEnabled()` , both return as `true`. It hurts my brain. Yet if I manually return true from this I can test the flows. ¯\_(ツ)_/¯

dallashuggins:
> re: never showing crypto and wires, I noticed this but I'm actually seeing this happen in `main` too. I thought it was just my environment. It's super weird because if I log `isEnabled` and `isEnabled.toLowerCase() === 'true'` in `get isWireEnabled()` , both return as `true`. It hurts my brain. Yet if I manually return true from this I can test the flows. ¯_(ツ)_/¯

Oh weird! Yeah I realized my environmental variables were set to true and haven't been seeing those are options, so it must've been an issue from a while back. Super confusing - going to look at that more.

smonn (approved):

jakerainis (approved):
Love seeing so much red. <3

afraser:
@dallashuggins ok so I've been trying to figure out the hook error (SUPER weird btw), and just noticed that I'm seeing that happen on `main` as well 😬.

dallashuggins:
> @dallashuggins ok so I've been trying to figure out the hook error (SUPER weird btw), and just noticed that I'm seeing that happen on `main` as well 😬.

Aw gotcha - thanks for looking into it! Makes sense to handle in a future PR.

dallashuggins (approved):
🔥

PR #412, opened 2022-04-11
Upgrade deps + update docs

smonn:
- Still on React v17 due to some incompatibilty issues with modals (@headlessui/react)
- This also means any dependency that needs React v18 is also "locked" until we can upgrade
- Add Scribe readme and move around/reword/update some docs


apps/cms/README.md
```diff
@@ -64,6 +64,24 @@ When you first spin up the app, you'll see only text rendered in place of images
...
+1. Navigate to the CMS and login
+1. Go to Settings > Webhooks
+1. Click on Create Webhook (the round plus button)
+1. Enter Scribe Webhook details and click save.
+
+| Field           | Value                                                                                                                           |
+| --------------- | ------------------------------------------------------------------------------------------------------------------------------- |
+| Name            | `Scribe` (or any name you prefer)                                                                                               |
+| Method          | `POST`                                                                                                                          |
+| URL             | `SCRIBE_URL` + `/webhooks/directus`                                                                                             |
```
dallashuggins, line 78: Would this be `http://localhost:[scribe port]/webhooks/directus` locally?

dallashuggins (approved):
Everything is working well locally 💯

apps/cms/README.md
smonn: That is correct

PR #416, opened 2022-04-11
Load font from Google Fonts to enable font optimization

smonn:
https://nextjs.org/docs/basic-features/font-optimization

> Automatic Webfont Optimization currently supports Google Fonts and Typekit with support for other font providers coming soon. We're also planning to add control over [loading strategies](https://github.com/vercel/next.js/issues/21555) and font-display values.


dallashuggins (approved):
Nice find! Fonts are displaying as they did before

tj-boyle (approved):

PR #418, opened 2022-04-12
Upgrade deps, use legacy-peer-deps

smonn:
- Set legacy-peer-deps=true in `.npmrc`
- Configure npm to use legacy-peer-deps in Dockerfiles

This _should_ ensure `package-lock.json` does not change if you forget to set `--legacy-peer-deps` when running `npm install` and also ensures you **don't** have to set it every time.

Reminder regarding React v18: still cannot bump the version until `@headlessui/react` officially supports it.


dallashuggins (approved):
Nice! Everything is still working well locally. I didn't test Docker however.

PR #419, opened 2022-04-13
Currency fixes

afraser:
This PR aims to simplify currency handling by only ever converting at:
 - **render time** — `<Currency>`, `<CurrencyInput>`, and `formatCurrency`
 - **payment time** — `PaymentsService` uses `convertToUSD` since all transactions are in settled in USDC

## CurrencyInput
This input accepts and returns values in USD cents now and does conversion internally.

Input is currently entered in USD (regardless of locale/currency) while displaying a localized price in the place of help-text (this is a temporary solution until we get designs). See The filter inputs video below.

Since all the conversion happens in one place, it should be easy to update this in the future to accept input in whatever currency we want. We'd simply load the reverse conversionRate from the `getCurrencyConversion` endpoint into context and get it through a hook. The sticky point here is making it clear to the user _depending on the context_ that all funds settle in USD so the value entered should probably be considered approximate.

## Bug fixes

There were several bugs in the auction bidding flow. Client side validation and bid initialization were broken and bids made in different currencies weren't being converted properly (12GBP was attempting to bid 12USD). We were also using react hooks on the backend so the add-to-pack endpoint was (thankfully) completely failing.

CleanShot 2022-04-19 at 08 56 40@2x

CleanShot 2022-04-19 at 08 58 20@2x

## Filter inputs

https://user-images.githubusercontent.com/1179515/163871518-afce0df3-30cb-4913-859e-b6abef01f5a7.mp4

## Notes
We often use `formatCurrency` because we need to pass a string into a translate function, but this requires a decent amount of boilerplate and can be (has been) misused. An alternative is to use <Trans> with <Currency> but this is ugly as hell. Better would be to write a helper function/component that does this for us while making the implementation more readable.


apps/web/src/components/bid-activity/sections/bid-activity-details.tsx
```diff
@@ -22,7 +24,11 @@ export default function BidActivityDetails({
         <h2 className={css.heading}>{content}</h2>
         <p className={css.date}>{date}</p>
       </div>
-      {amount && <p className={css.amount}>{amount}</p>}
+      {amount && (
+        <p className={css.amount}>
+          <Currency value={amount} />
+        </p>
+      )}
```
smonn, line 31: Should this be changed to handle the case of `amount` being `0` (zero)? Or is it intentional to not show this if the amount is zero?

```suggestion
      {typeof amount === 'number' && (
        <p className={css.amount}>
          <Currency value={amount} />
        </p>
      )}
```

apps/web/src/contexts/payment-context.tsx
```diff
@@ -151,7 +146,7 @@ export function usePaymentProvider({
     [t]
   )
   const [formErrors, setFormErrors] = useState<FormValidation>()
-  const [price, setPrice] = useState<string | null>()
+  const [price, setPrice] = useState<number>(0)
```
smonn, line 149: If the initial value matches the desired type of the `useState` hook, no need to set it explicitly.

```suggestion
  const [price, setPrice] = useState(0)
```


We're not very consistent with this in our code, but trying to address that.

smonn (commented):
Made an initial review pass. Gonna test it out locally too, but left a few comments here to start.

apps/web/src/components/bid-activity/sections/bid-activity-details.tsx
afraser: I'm 99% sure that this is just one of those cases that doesn't matter because in this context (auction bids) there should never be a bid of zero.

afraser:
Dallas caught a bug when doing a crypto checkout that I think is in a file I had some suspicions about. Meanwhile I found another place. Looking at these two things right now:

payments.service.ts: createTransferPayment()
crypto-form-wc.tsx: handleWalletConnectPurchase()

patnealcodes (commented):
Awesome stuff! lgtm

apps/web/src/components/purchase-form/crypto/sections/crypto-form-wc.tsx
```diff
@@ -125,7 +101,7 @@ export default function CryptoFormWalletConnect({
     if (connector) {
       setLoadingText(t('common:statuses.Connected to Wallet'))
       const assetTx = await algorand.makeAssetTransferTransaction({
-        amount: usdcPriceInt * 10_000,
+        amount: price,
```
afraser, line 104: @smonn does this seem right to you? we're transferring usdc here, not micro-algos.

afraser:
Some before/after images for updated parts of crypto flow:

wc before
wc after

before
after

apps/web/src/components/purchase-form/crypto/sections/crypto-form-wc.tsx
```diff
@@ -93,27 +81,15 @@ export default function CryptoFormWalletConnect({
...
-      usdcAsset.amount,
-      usdcAsset.decimals,
-      currency,
-      conversionRate
-    )
-
-    const usdcBalanceInt = formatFloatToInt(usdcBalance)
-
-    const usdcPriceInt = formatFloatToInt(price)
-    const currencyPriceInt = formatFloatToInt(price, currency, conversionRate)
```
afraser, line 108: This was kinda weird that we're converting an int to a float and back to an int again. It'd make sense if we were dealing with varying currencies/coins, but here we're only interacting with USDC which appears to be returned in cents. Lucky us.

dallashuggins (approved):
Everything has been working well. The only issue is still just with crypto payments, and it seems to be on Circle's side, but was working earlier. Great improvements!

apps/web/src/components/purchase-form/crypto/sections/crypto-form-wc.tsx
afraser: This was unfortunately not terribly helpful: [`makeAssetTransferTxnWithSuggestedParamsFromObject`](https://algorand.github.io/js-algorand-sdk/modules.html#makeAssetTransferTxnWithSuggestedParamsFromObject)

smonn: Short version: assuming `price` here is in the correct denomination, yes this should be fine. 

Slightly longer version: USDCa is defined with 6 decimals ([see TestNet version](https://testnet.algoexplorer.io/asset/10458941) or [MainNet version](https://algoexplorer.io/asset/31566704)). This means `amount` here would be $1 = 1,000,000. It's similar to microAlgos in that sense; when sending a payment in ALGO you specify it as microAlgos where 1 ALGO = 1,000,000 microAlgos. I don't know how/where `price` came from so I can't give a straight answer, but I'd verify that you're providing "microUSDCa" here. So if you're expecting $10, then the value should be 10,000,000.

afraser: Ah ok! So price here (and everywhere now) is in cents so yeah (price / 100) * 1,000,000 would give us that 10_000 multiplier. I'll put that back in and we should be good to merge.

PR #421, opened 2022-04-14
seed script refactors

tj-boyle:
- refactoring seeds into reusable functions
- switching to directus sdk
- splitting dev and prod for easy theoretical prod setup

### DIFF
--- a/apps/cms/scripts/directus.mjs
+++ b/apps/cms/scripts/directus.mjs
@@ -22,7 +22,8 @@ export async function setupSDK(config) {
         authenticated = true;
       })
       .catch(() => {
-        window.alert('Invalid credentials');
+        console.error('Invalid Directus credentials');
+        process.exit(1);
       });
   }
 
--- a/apps/cms/scripts/directus.mjs
+++ b/apps/cms/scripts/directus.mjs
@@ -1,12 +1,70 @@
 import { Directus } from '@directus/sdk'
 
-export async function configureDirectus() {
-  const client = new Directus(process.env.PUBLIC_URL)
+export async function setupSDK(config) {
+  const directus = new Directus(process.env.PUBLIC_URL);
 
-  await client.auth.login({
-    email: process.env.ADMIN_EMAIL,
-    password: process.env.ADMIN_PASSWORD,
-  })
+  // But, we need to authenticate if data is private
+  let authenticated = false;
 
-  return client
+  // Try to authenticate with token if exists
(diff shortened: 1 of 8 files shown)


apps/cms/scripts/seed-dev.mjs
```diff
@@ -146,82 +116,70 @@ async function main() {
...
-      const previewImage = await makeImage({
-        directus,
-        text: item.translations[0].title,
-        color: '#000000',
-        borderColor: rarity?.color || '#000000',
-        backgroundColor: randColor(),
-      })
-
-      item.preview_image = previewImage.id
+  const buildList = Factory.buildList('collectible', numCollectibles)
```
tj-boyle, line 145: I switched this to a pure synchronous for loop for two reasons

1. As `Promise.all`, the calls to `createAssetRecords` in `makeImage` queue up and then get called all at the same time. With larger multipliers this was bombarding the CMS with calls and causing service unavailable errors. By making it purely synchronous, it is much easier on the CMS
2. Additionally with larger multipliers, this process can hang for a while. I added a console log for each image being processed so you can have an idea that it's actually doing something and not hanging indefinitely

apps/cms/scripts/directus.mjs
```diff
@@ -1,12 +1,70 @@
...
+
+  // Let's login in case we don't have token or it is invalid / expired
+  while (!authenticated) {
+    await directus.auth
+      .login(config)
+      .then(() => {
+        authenticated = true;
+      })
+      .catch(() => {
+        window.alert('Invalid credentials');
```
smonn, line 25: This won't work, we're running this in a Node.js context. Something like this might be better:

```suggestion
        console.error('Invalid Directus credentials');
        process.exit(1);
```

apps/cms/scripts/directus.mjs
```diff
@@ -1,12 +1,70 @@
...
+
+export async function setAccessToken(directus) {
+  const me = await directus.users.me.read();
+
+  await directus.users.updateOne(me.id, {
+    token: process.env.ADMIN_ACCESS_TOKEN,
+  });
+}
+
+export async function setFilePermission(directus) {
```
smonn, line 40: _Technically_, we don't need the file permissions to be public if you use the GCP bucket and its read-access is public. But could be good to keep it around?

smonn (approved):
Minor comments, looks good!

PR #422, opened 2022-04-14
lang redirect, ltr->rtl change

tj-boyle:
was missing redirect to lang on load, and document direction change if rtl language selected

### DIFF
--- a/apps/web/src/pages/_app.tsx
+++ b/apps/web/src/pages/_app.tsx
@@ -1,3 +1,4 @@
+import { RTL_LANGUAGES } from '@algomart/schemas'
 import type { AppProps } from 'next/app'
 import { useRouter } from 'next/router'
 import { useEffect } from 'react'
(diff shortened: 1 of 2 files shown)


smonn (approved):

afraser (approved):
Another nice-to-have would be to update `ChevronRightIcon` in components/step-links.tsx to use `ChevronLeftIcon`

CleanShot 2022-04-14 at 16 23 55@2x

afraser:
one other thing we _might_ want to do is update this page to use flex with reverse-row and ChevronLeftIcon

CleanShot 2022-04-14 at 16 25 13@2x


I've never done RTL before though so I wonder if there's a more general way of doing this (ie: we could use tailwind variables to set the default flex direction based on rtl/ltr (granted I think we'd also need to add a class to html).

tj-boyle:
Yeah @afraser there's definitely a larger effort to be done to fully support RTL languages, this is just the first pass to hook language changing up to the browser's rtl/ltr functionality

PR #423, opened 2022-04-14
Fix: replace references to environment on client with useConfig

afraser:


apps/web/src/hooks/use-export-collectible.ts
```diff
@@ -26,6 +24,12 @@ export function useExportCollectible(passphrase: string) {
   const [selectedAccount, selectAccount] = useState('')
   const [exportStatus, setExportStatus] = useState<ExportStatus>('idle')
   const connectorReference = useRef<IConnector>()
+  const Environment = useConfig()
+
+  const algorand = useMemo(
```
afraser, line 29: I didn't test this but figure we can cross this bridge when we come back to it. LMK if anything seems off though.

apps/web/src/hooks/use-export-collectible.ts
```diff
@@ -26,6 +24,12 @@ export function useExportCollectible(passphrase: string) {
   const [selectedAccount, selectAccount] = useState('')
   const [exportStatus, setExportStatus] = useState<ExportStatus>('idle')
   const connectorReference = useRef<IConnector>()
+  const Environment = useConfig()
```
patnealcodes, line 27: Might it make sense to name this 'Config' or something other than 'Environment' since it's on the client side? Just to avoid confusion with the server>client environment variables?

apps/web/src/hooks/use-import-collectible.ts
```diff
@@ -1,6 +1,6 @@
 import { useCallback, useMemo, useRef, useState } from 'react'
 
-import { Environment } from '../environment'
+import { useConfig } from './use-config'
```
patnealcodes, line 3: 🧠

patnealcodes (approved):
lgtm

apps/web/src/hooks/use-export-collectible.ts
afraser: yeah you're totally right, I guess I was being lazy about updating these

smonn (approved):

jakerainis (approved):

PR #424, opened 2022-04-14
Algorand shared lib to implement various ARCs

smonn:
Remaining todos:

- [x] Helpers to construct ARC-3 metadata
- [x] Refactor existing code to utilize these helpers (this is a precursor to queuing txns)

This implements the following ARCs:

- [ARC-1](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0001.md): Wallet signing API. Technically for noncustodial wallets, but helpful for serializing txns and ensure compatibility with WalletConnect etc when requiring signatures from both custodial and noncustodial wallets.
- [ARC-2](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0002.md): Transaction note field. Adds extra metadata to the note field for transactions. No strong opinions on what we store, but setting this up as a start.
- [ARC-3](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0003.md): NFTs. Technically we already support this, just codifying it a bit more and ensuring we have it all in one place. May need some tweaks with ARC-18 and also add helpers to construct the metadata JSON.
- [ARC-5](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0005.md): Similar to ARC-1, but technically we follow this too.
- [ARC-18](https://github.com/algorandfoundation/ARCs/pull/70): Royalty contract (WIP). This is just partial support at the moment (from the ASA perspective, implementing smart contracts later).

Other ARCs to be considered:

- [ARC-69](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0069.md). Not sure yet if this will work with ARC-3?
- Others...?

One unchanged file is the public `AlgorandAdapter` used in the web app. Not sure yet how we want to handle that since it requires a public indexer/algod at the moment.


jakerainis:
Keeping an eye on this. Would love to walk through it together once you feel it's ready for review.

apps/scribe/src/tasks/configure-tasks.ts
```diff
@@ -48,6 +49,17 @@ export function configureTasks(app: FastifyInstanceWithScheduler) {
     )
   )
 
+  app.scheduler.addSimpleIntervalJob(
+    new SimpleIntervalJob(
+      { seconds: 5 },
```
smonn, line 54: Not sure about the duration for this and the previous one, let's discuss!

libs/shared/adapters/src/lib/cms-cache-adapter.ts
```diff
@@ -432,7 +431,7 @@ export class CMSCacheAdapter {
   }
 
   async findAllPacksAuctionCompletion(
-    startDate,
+    startDate: Date,
```
smonn, line 434: A few missing typings in this file

libs/shared/adapters/src/lib/cms-cache-adapter.ts
```diff
@@ -516,28 +515,11 @@ export class CMSCacheAdapter {
     return data
   }
 
-  async findPacksByType(
```
smonn, line 519: Found no usages of this method.

libs/shared/adapters/src/lib/cms-cache-adapter.ts
```diff
@@ -794,25 +776,6 @@ export class CMSCacheAdapter {
     return result
   }
 
-  private async findSets(query: ItemQuery = {}, trx?: Transaction) {
```
smonn, line 797: Was not being used.

libs/shared/adapters/src/lib/cms-cache-adapter.ts
```diff
@@ -887,13 +850,17 @@ export class CMSCacheAdapter {
...
 
-    return queryBuild.page(
-      query.page - 1 || 0,
-      query.limit != -1 ? query.limit : Number.MAX_SAFE_INTEGER
-    )
+    const page = query.page - 1 || 0
+    const pageSize =
+      query.limit != -1 ? query.limit || 10 : Number.MAX_SAFE_INTEGER
+
+    return queryBuild.page(page, pageSize)
```
smonn, line 857: Not sure if this is the correct/expected behavior, but was seeing warning logs because the old version passed `undefined` as the `pageSize`.

apps/web/src/templates/nft-template.tsx
```diff
@@ -35,7 +35,6 @@ function getTransferrableStatus(
 ) {
   if (!currentUserAddress) return 'noUser'
   if (collectible.currentOwnerAddress !== currentUserAddress) return 'notOwner'
-  if (collectible.isFrozen) return 'frozen'
```
smonn, line 38: Since we set `defaultFrozen` to true again, this is no longer preventing us from doing exports. The export/import code is also updated to rely on clawbacks.

dallashuggins (approved):
Looks great! 🔥 Everything is working great when I try it on my own as well. Thanks for the walk through!

PR #425, opened 2022-04-15
Stripe Identity integration

dallashuggins:
### Changes
- Stripe integration
  - Added `/profile/verify` page where verification can be initiated using Stripe Identity
  - New environmental variables. For web, the Stripe public API key. For api and scribe, the Stripe API secret. 
  - API endpoints for creating a verification session and getting a verification session for a user.
  - User table has new `verificationId` column for storing the current verification session. 
  - Stripe SDK


dallashuggins:
Closing since Stripe Identity doesn't do ongoing monitoring, but this could be useful down the road.

PR #435, opened 2022-04-20
Fixes #426 - Replace crypto.randomUUID with something simpler

### LINKED ISSUE #426: Unhandled Runtime Error
## Overview

On mobile only, an error pops up for me that says 
TypeError: crypto.randomUUID is not a function. (In 'crypto.randomUUID()', 'crypto.randomUUID' is undefined)

## To reproduce

Only, when I load up mobile

## Expected behavior

The webpage loads

## Screenshots

image

## Desktop (if applicable):

Linux, Firefox

## Mobile (if applicable):

IOS, Chrome/Safari

## Additional context
I checked the function mentioned in the error does at least work correctly.
```Welcome to Node.js v16.13.1.
Type ".help" for more information.
> crypto.randomUUID()
'9befc171-d31a-4bf4-ba53-e5784960a897'
```

DeadZen:
Submitted is a pull request for issue #426 reported by me.
This fix might work better for backwards compatibility among mobile users.

### DIFF
--- a/apps/web/src/components/select/select.tsx
+++ b/apps/web/src/components/select/select.tsx
@@ -44,7 +44,7 @@ export default function Select({
   value,
   Icon,
 }: SelectProps) {
-  const _id = id ?? crypto.randomUUID()
+  const _id = id ?? '_' + Math.random().toString(36).substr(2, 9)
   const [internalValue, setInternalValue] = useState(
     defaultValue || (options?.length ? options[0].value : '')
   )
--- a/apps/web/src/components/text-input/text-input.tsx
+++ b/apps/web/src/components/text-input/text-input.tsx
@@ -24,7 +24,7 @@ export default function TextInput({
   ...props
 }: TextInputProps &
   DetailedHTMLProps<InputHTMLAttributes<HTMLInputElement>, HTMLInputElement>) {
-  const _id = id ?? crypto.randomUUID()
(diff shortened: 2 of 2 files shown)


smonn (commented):
I'm okay with this to avoid using the in-browser Crypto API. Not sure if this is _good enough_ randomization for these IDs though. @afraser?

DeadZen:
@smonn @afraser I pushed again to increase the "good enough" aspects, its now prefixed with the current datetime to reduce the likelihood of a randomized id collision

apps/web/src/components/select/select.tsx
```diff
@@ -44,7 +44,7 @@ export default function Select({
   value,
   Icon,
 }: SelectProps) {
-  const _id = id ?? crypto.randomUUID()
+  const _id = id ?? '_' + Date.now().toString(36) + Math.random().toString(36).substr(2, 9)
```
smonn, line 47: Let's switch to using `uuid` (it's already installed and ready to use). First import it at the top:

```ts
import { v4 as uuid } from 'uuid'
```

Then use it like so:

```suggestion
  const _id = id ?? uuid()
```

smonn (commented):
Hey @DeadZen, just a minor change and then we should be good!

smonn (approved):

PR #448, opened 2022-04-28
currency fixes

afraser:
## Fixes
- Setting user preference for currency & language was broken
  - Currency appeared to work before because it was reading from the cookie
  - Language worked... some of the time? (eg: changing language and then clicking on the avatar icon wouldn't respect the new language selection when routing)
- Hydration issue on static pages when displaying non-default currency
- Filtering on the releases page was broken for currencies other than USD

## Refactoring
- cookie updates are surfaced across the app without needing a refresh
- `updateCurrency` and `updateLanguage` are now centralized and need not be redefined in multiple places
- user currency is now preferred over cookie currency (used for guests)

## My profile
- validation removed from select inputs since we should never be rendering invalid options in them.
- save/cancel flow removed for simplicity and to match theme selection.

https://user-images.githubusercontent.com/1179515/165985567-45c541c5-842d-4127-9a30-28bc213afc86.mp4

https://user-images.githubusercontent.com/1179515/165985582-fee9c08f-c554-4469-802f-a8310be17dc2.mp4

☝🏼 in this video you can see the filtering bug for currencies other than USD... turns out the server was converting the value even though we pass it in USD cents.

Here it is fixed:

https://user-images.githubusercontent.com/1179515/165991406-969f70f4-26b7-4eff-a1a3-fb7c63a3be6d.mp4


apps/web/src/components/app-footer/sections/app-footer-currency.tsx
```diff
@@ -1,84 +1,9 @@
-import { useRouter } from 'next/router'
-import useTranslation from 'next-translate/useTranslation'
-import React, { useCallback, useEffect, useMemo, useState } from 'react'
+import React from 'react'
```
tj-boyle, line 1:
```suggestion
```

don't think this is necessary

apps/web/src/contexts/currency-context.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export const CurrencyProvider = ({
+  children,
+}: {
+  children: React.ReactNode
+}) => {
+  const auth = useAuth(false)
+
+  // We need to store the cookie value in state in order to force re-render on change.
+  const [cookie, setCookie] = useState(getCookie(CURRENCY_COOKIE))
```
tj-boyle, line 29: That explains a LOT

apps/web/src/contexts/currency-context.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const parsedCookie = cookie && cookie !== 'null' ? cookie : null
+
+  const setCurrency = async (currency: string) => {
+    // always update cookie even though it's only used when not logged in
+    setCookie(currency)
```
tj-boyle, line 34:
```suggestion
    setCookie(currency)
```

don't think this line is necessary, the next line does this job

tj-boyle (approved):
Love this, thanks for handling my awfulness

apps/web/src/components/app-footer/sections/app-footer-currency.tsx
patnealcodes: are you saying that this is an over-Reaction?

apps/web/src/components/currency/index.tsx
```diff
@@ -22,8 +25,22 @@ export default function Currency({
...
+    locale = DEFAULT_LOCALE
+  }
+
+  useEffect(() => {
+    /*
+     * This addresses an issue where the user's selected currency is not
+     * available at hydration time, causing a discrepancy between server and client
+     */
+    setHydrated(true)
+  }, [])
```
smonn, line 43: 🎉

apps/web/src/components/profile/my-profile-currency.tsx
```diff
@@ -1,92 +1,36 @@
...
-        { pathname: router.pathname, query: router.query },
-        router.asPath
-      )
-
-      return
+      updateCurrency(currency).then((success) => {
+        if (success) {
+          setError('')
+          setUpdateSuccess(true)
+          setTimeout(() => setUpdateSuccess(false), 3000)
```
smonn, line 25: Can we add a comment here? Not sure I follow what's going on/why this is needed.

apps/web/src/components/profile/my-profile-language.tsx
```diff
@@ -1,98 +1,36 @@
...
-        router.asPath,
-        { locale: body?.language }
-      )
-
-      return
+      updateLanguage(language).then((success) => {
+        if (success) {
+          setError('')
+          setUpdateSuccess(true)
+          setTimeout(() => setUpdateSuccess(false), 3000)
```
smonn, line 25: Same.

apps/web/src/contexts/currency-context.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export const CurrencyProvider = ({
+  children,
+}: {
+  children: React.ReactNode
+}) => {
+  const auth = useAuth(false)
+
+  // We need to store the cookie value in state in order to force re-render on change.
+  const [cookie, setCookie] = useState(getCookie(CURRENCY_COOKIE))
```
smonn, line 29: Is this safe to be called here? Since it would be called both on the server and in the browser (`js-cookie` is a browser library).

apps/web/src/contexts/currency-context.tsx
```diff
@@ -0,0 +1,53 @@
...
+      const success = await AuthService.instance.updateCurrency(currency)
+      await auth.reloadProfile()
+      return success
+    }
+    return true
+  }
+  const value = {
+    currency: auth?.user?.currency || parsedCookie || DEFAULT_CURRENCY,
+    updateCurrency,
+  }
```
smonn, line 46: Not to overdo hooks, but would a `useMemo` be helpful here to avoid re-renders? Similarly, should `updateCurrency` be wrapped in `useCallback` (or defined within the `useMemo` hook)?

apps/web/src/contexts/language-context.tsx
```diff
@@ -0,0 +1,101 @@
...
+    // 3rd router locale
+    // 4th browser language
+    // 5th english
+    setLanguage(getSupportedLanguage(parsedCookie || fallbackLanguage))
+  }, [fallbackLanguage])
+
+  const value = {
+    language,
+    updateLanguage,
+  }
```
smonn, line 94: Same as previous comment.

apps/web/src/contexts/payment-context.tsx
```diff
@@ -718,7 +718,7 @@ export function usePaymentProvider({
 
   useEffect(() => {
     setPrice(release?.type === PackType.Auction ? bid : release?.price || 0)
-  }, [currency, bid]) // eslint-disable-line react-hooks/exhaustive-deps
+  }, [bid]) // eslint-disable-line react-hooks/exhaustive-deps
```
smonn, line 721: I realize the eslint comment precedes these changes, but curious if it's still needed? Obviously the `currency` variable is not read in here, but `release` is and not listed. Not sure what the original reasoning was for disabling the rule.

smonn (commented):
Some comments I'd like to get clarified. But looks good overall!

apps/web/src/contexts/payment-context.tsx
patnealcodes: My assumption here is that we wouldn't want to re-run the `setPrice` function when release value is changed, only when the bid is

apps/web/src/components/profile/my-profile-currency.tsx
afraser: you bet, I'll also try to come up with a better name for this. Took me a minute to figure out what the heck it was doing too.

Basically it's whether or not to show a success message after saving. So we show the message and then hide after 3s.

apps/web/src/contexts/currency-context.tsx
afraser: ooh I wonder if we can fix all our hydration issues by moving those checks here.

apps/web/src/contexts/language-context.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  if (SUPPORTED_LANGUAGES.includes(lang)) return lang
+  const [langPrefix] = lang.split('-')
+  return SUPPORTED_LANGUAGES.includes(langPrefix) ? langPrefix : DEFAULT_LANG
+}
+
+export const LanguageProvider = ({
+  children,
+}: {
+  children: React.ReactNode
```
jakerainis, line 35: Just for consistency, could import `ReactNode` from the already-present `react` import.

jakerainis (approved):
Nice work!

apps/web/src/components/currency/index.tsx
afraser: Ended up removing this in favor of a fix in currency-context.

apps/web/src/contexts/payment-context.tsx
afraser: went ahead and removed the disable in favor of explicitly listing the deps. My guess is they were omitted because they should never change.

patnealcodes (approved):
looks great

PR #449, opened 2022-04-28
create collectible listings table

tj-boyle:
Table to hold information on a listing for querying and display purposes

CollectibleListings
- id - uuid
- collectibleId - FK Collectible.id
- price - number
- type - [purchase, auction]
- expiresAt - Date
- createdAt - Date
- updatedAt - Date

Could potentially expand this table down the line to include things like:
- Flag for archived or not (current strategy is to delete row when listing removed or completed)
- Flag for status, potentially to hold an "in flight" state if someone else is already in the purchase flow for the Listing

### DIFF
--- a/apps/scribe/src/migrations/20220428190347_collectible_listings.ts
+++ b/apps/scribe/src/migrations/20220428190347_collectible_listings.ts
@@ -1,0 +1,21 @@
+import { Knex } from 'knex'
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.createTable('CollectibleListings', (table) => {
+    table.uuid('id').primary()
+    table
+      .uuid('collectibleId')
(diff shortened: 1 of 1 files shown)


dallashuggins (approved):
💯

dallashuggins:
>     Flag for archived or not (current strategy is to delete row when listing removed or completed)
>     Flag for status, potentially to hold an "in flight" state if someone else is already in the purchase flow for the Listing

Maybe archived could be a type in addition to purchase/auction? Assuming the user couldn't ever relist an archived listing.
Status could be helpful, though we could also convey on the release page. If we updated the status anytime a purchase was started, we'd have to make sure that was updated whenever they left, and the data could be inaccurate on the releases page depending on when the purchase was started. 🤔

PR #450, opened 2022-04-29
Product Search

tj-boyle:
- [x] Create collectibleListings table
- [x] Modify path to be "Browse" with "Products" context
- [x] Make boilerplate Product UI component
- [x] Make new search Products endpoint and service
- [x] Make new Product model that is derived from Pack Templates and Collectible Templates + Collectible Listings
- [x] DB query
  - [x] Knex join for combining CmsCachePackTemplates and CmsCacheCollectibleTemplates that have a Collectible 
  - [ ] New sql filter function for searching the previously mentioned Knex joinListing
- [x] Post DB query
  - [x] Functions for getting extra Pack Template data post-filter (auction information....etc)
  - [ ] Functions for getting extra Collectible Template data post-filter (tbd)
- [ ] Modify Product component to expect new Product model
- [ ] Cleanup now unused PackTemplate filter code, etc

**A note on the knex query:**
It's a union, mapping columns from CmsCachePackTemplate and CollectibleListings (joined Collectible joined CmsCacheCollectibleTemplate) to a common Product model.

The idea is that in the UI the card would represent a Collectible Template that has at least 1 Collectible for sale, and a user would click through to see a page with all that template's listings.
Therefore, the query groups CmsCacheCollectibleTemplates, only showing the first cheapest listing for the given template (cheapest because that's what the price filter will be using for it's range). 

It puts these alongside CmsCachePackTemplates, morphing them both into a common derived model, _**Products.**_

This was a first pass, it's possible there's a more performant way to do the query.

**Another way?**
We could potentially just make an actual "Products" table, that we update when a Pack Template or a Collectible Listing gets made/updated. This would involve changing sync flows and service flows, and would mean we'd likely have 2 sources of truth for some data, which would not be ideal. This may be worth digging into more if the query becomes non-performant or another usecase arises


libs/shared/adapters/src/lib/cms-cache-adapter.ts
```diff
@@ -733,6 +734,52 @@ export class CMSCacheAdapter {
     return result
   }
 
+  async findAllProducts(query: ItemQuery = {}, trx?: Transaction) {
+    const queryBuild =
+      CMSCachePackTemplateModel.query(trx)
```
tj-boyle, line 739: 💀

PR #457, opened 2022-05-02
upgrade dependencies

smonn:
Upgrade to latest version of most dependencies. Jest v28 is not yet compatible so had to hold that off for now. Fastify moved to a scoped model for its modules, so that required some code changes. React Dropzone changed its API slightly. Other than that, everything seems to work the same as before.


jakerainis (approved):
Haven't tested but I trust ya! Thanks for doing this.

dallashuggins (approved):
Everything is working well 💯

PR #466, opened 2022-05-10
Added some notes to `README`’s and `.env.sample`’s

patnealcodes:
Motivation: This PR updates some notes in the README files and adds some examples to the `.env` sample files to add more clarity around the setup process

### DIFF
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -1,27 +1,29 @@
 # API configuration
-HOST=
-PORT=
-API_KEY=
+HOST= ;e.g. localhost
+PORT= ;e.g. 3001
+API_KEY= ;any string, e.g. some_apy_key
 
 # Postgres configuration
-DATABASE_URL=
+DATABASE_URL= ;match the DB you’ve created, e.g. postgres://postgres:postgres@localhost:5432/algomart
 DATABASE_SCHEMA=public
 
 # Algod configuration
-ALGOD_SERVER=
-ALGOD_PORT=
-ALGOD_TOKEN=
-ALGOD_ENV=testnet
+ALGOD_SERVER= ;e.g. http://localhost
+ALGOD_PORT= ;e.g. 4001
+ALGOD_TOKEN= ;e.g. aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+ALGOD_ENV= ;e.g. testnet
 
-# CMS configuration (CMS_ACCESS_TOKEN should match the Admin User > Admin Options > Token field in the CMS UI)
-CMS_ACCESS_TOKEN=
+# CMS configuration
+
+CMS_ACCESS_TOKEN= ;any string, should match the Admin User > Admin Options > Token field in the CMS UI
 CMS_URL=http://localhost:8055
 
 # Web app configuration
 WEB_URL=http://localhost:3000
 
 # Circle API configuration
-CIRCLE_API_KEY=
+
+CIRCLE_API_KEY= ;create a free circle account to generate a key
 CIRCLE_URL=https://api-sandbox.circle.com
 
 # Email configuration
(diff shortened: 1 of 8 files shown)


apps/api/.env.example
```diff
@@ -1,27 +1,29 @@
...
 
 # Algod configuration
-ALGOD_SERVER=
-ALGOD_PORT=
-ALGOD_TOKEN=
-ALGOD_ENV=testnet
+ALGOD_SERVER= ;e.g. http://localhost
+ALGOD_PORT= ;e.g. 4001
+ALGOD_TOKEN= ;e.g. aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+ALGOD_ENV= ;e.g. testnet
```
brandonaaskov, line 14: Someone new to this might not realize that something like that ALGOD_TOKEN is actually valid (for dev). Is there a downside to just defining the dev defaults as opposed to showing the `;e.g.`?

apps/api/.env.example
```diff
@@ -1,27 +1,29 @@
...
 CMS_URL=http://localhost:8055
 
 # Web app configuration
 WEB_URL=http://localhost:3000
 
 # Circle API configuration
-CIRCLE_API_KEY=
+
+CIRCLE_API_KEY= ;create a free circle account to generate a key
 CIRCLE_URL=https://api-sandbox.circle.com
```
brandonaaskov, line 27: I guess we are already using dev defaults in some places. That's preferred IMO but that might just be a taste thing.

apps/api/.env.example
```diff
@@ -48,16 +50,19 @@ SMTP_USER=
...
 
 # Mnemonic for master funding & signing account
-FUNDING_MNEMONIC=
+
+FUNDING_MNEMONIC= ;from the sandbox account you created (check the README.md)
 
 # Used for passphrase encryption
-SECRET=
+
+SECRET= ;any string
```
brandonaaskov, line 58: Should we recommend a min length or maybe something obfuscated? This one seems important for something like that.

apps/api/.env.example
```diff
@@ -48,16 +50,19 @@ SMTP_USER=
...
 
 # Used to generate blockchain assets
 CREATOR_PASSPHRASE=
 
 # Currency for application - needs to match the web env file and the CMS currency field.
-CURRENCY=
+
+CURRENCY= ;e.g. USD
```
brandonaaskov, line 65: What would one or two other options look like? Euros, pesos?

apps/api/README.md
```diff
@@ -35,7 +37,23 @@ To create an account:
...
+./sandbox goal account list
+```
+
+- You should see several accounts, some with _microAlgos_, and the account with the address you created above likely having 0. You can run the following command to fund the account you just created with microAlgos
+
+```bash
+./sandbox goal clerk send -a <AMOUNT> -f <FROM_ADDRESS> -t <TO_ADDRESS>
+
+# Note: Something like 1000000 microAlgos as the amount should be more than plenty to cover transactions
+```
```
brandonaaskov, line 56: This is a very helpful addition, thank you.

apps/cms/README.md
```diff
@@ -27,9 +27,11 @@ Start the server locally:
 nx serve cms
 ```
 
-Optionally, you can seed the database (while the server is running) to generate dummy content:
+Optionally, you can seed the database to generate dummy content.
 
 ```bash
+# While the server is runnning:
+
```
brandonaaskov, line 34: That's a subtle change but a good one. Most people go straight to the code and that's a good reminder 👍

brandonaaskov (commented):
I'm not close enough to the daily usage to approve or disapprove, just leaving some comments :)

apps/api/.env.example
patnealcodes: Not that I'm aware and I'm happy to update if there are no objections

apps/api/.env.example
jakerainis: The `CIRCLE_API_KEY` has to be unique per project so we can't default one here, might suggest linking directly to the signup page where a sandbox key can be provisioned freely for development purposes (https://my-sandbox.circle.com/signup)

apps/api/.env.example
jakerainis: One thing we frequently use to generate keys on the fly is:
```sh
$ openssl rand -hex 32 | pbcopy
```

Which will generate something like: `5f6a12942bb2c44eb70cf1b71ef1c211cc2d966981f76e6315bd086fe54cac16`. Could be worth mentioning at the top of the .env file — or not. Just a good thing to be aware of!

jakerainis (approved):
Good stuff! Thank you for adding to this. It's always helpful to have a fresh set of eyes!

apps/api/.env.example
patnealcodes: Oh that's awesome, I'll definitely add that in!

